PR #5, opened 2023-10-15
Cutscene skips for Mar Sara, Whispers, Piercing, and Smash+Grab

EnvyDragon:
Did the following:
-Added skip cutscene flags 
-Added disable-forced-camera-movement flags
-Removed a bunch of hint bubbles
-Removed a bunch of UI flashes
On the following maps:
-The Outlaws
-Zero Hour
-Smash and Grab
-Whispers of Doom
-Piercing the Shroud


Ziktofel:
I'd change the command to `?CutsceneAutoskip true|false` (use an argument).

EnvyDragon:
@Ziktofel 

> I'd change the command to `?CutsceneAutoskip true|false` (use an argument).

That's not ideal for this use-case.

The reason this command exists is so that a player can download these versions of the maps and use cutscene skips without needing the in-development client or yaml settings; in theory, you could run these on 0.4.3 or earlier and still save time on missions that make you wait.

Because cutscene skip preference isn't saved between missions this way, we want the argument to be as short as possible, so it's easy for players to type it in at the start of every single mission including restarts.

`?nocut` is 6 characters, which is already more than I'd like. `?CutsceneAutoskip true` is too many.

PR #6, opened 2023-10-15
Improved logic and added all Artifact missions

EnvyDragon:
Double-checked that all tips were removed and added some temp always-false variables that may eventually be replaced by actual archipelago flags on already-done maps.

Also added checks for skip-cinematics/disable-forced-movement flags, and removed tip buttons, from the following maps:
-The Dig
-The Moebius Factor
-Supernova
-Maw of the Void


Ziktofel:
you're setting vars like this in every map:
```
bool gv_temp_gameplay_skip_flag;
bool gv_temp_keep_transmissions_flag;
bool gv_temp_skip_victory_flag;
```
shouldn't these go into the Triggers mod?

PR #14, opened 2023-11-15
sc2: Fixing some incorrect unlocks for Blood Hunter, Energizer, and Protoss Air Upgrades

MatthewMarinets:
See issues on the main AP repo, [#90 - energizer](https://github.com/Ziktofel/Archipelago/issues/90), [#95 - Blood Hunter](https://github.com/Ziktofel/Archipelago/issues/95), and [#88 - stargate unlock from air upgrades](https://github.com/Ziktofel/Archipelago/issues/88).

I tested the fixes by manually re-applying them to the triggers file after I did a `/download_data`, and checking I could build energizers and blood hunters with the items in my inventory. I did _not_ do extra checking for the stargate, though I still tested things loaded with the stargate changes applied to the trigger file.

I noticed when I copy-pasted the whole file that the triggers failed to initialize; I'm not sure why, I think it's related to other issues that are in the repo but aren't in the published triggers file.


Ziktofel:
Picked just unit fixes. The Stargate is for being able to build Cybercore

PR #18, opened 2023-11-17
The Infinite Cycle Fix

MadiMadsen:
The second and third bonus objectives were 1 value higher then they should be causing the 3rd location to be sent incorrectly for the second. and no location to be sent  for the third.


MadiMadsen:
https://github.com/Ziktofel/Archipelago/issues/87#issue-1993859958 will fix this issue

Ziktofel (approved):

PR #21, opened 2023-11-22
Zealot Whirlwind turned into an upgrade.

MadiMadsen:
Zealot Whirlwind is now an upgrade that can be researched is currently on by default via a trigger in the trigger mod


Ziktofel (changes requested):
Missing some trigger-related stuff (galaxy file update, comment string to let know that's an un-nerf, etc

PR #25, opened 2023-11-25
Supplicant upgraded added

MadiMadsen:
shield go from 125 to 150 and shield armor gets a +2! is on by default.


MadiMadsen:
Also should note the other ability that normally gives +25 shield by itself will be reworked into a new ability in the future once supplicant upgrades are worked on

Ziktofel:
the +25 shield is another upgrade trait now:
```xml
    <CUpgrade id="AP_AlarakSupplicantMaxShields">
        <Flags index="TechTreeCheat" value="1"/>
        <Icon value="Assets\Textures\btn-upgrade-protoss-alarak-supplicantextrashields.dds"/>
        <Alert value="ResearchComplete"/>
        <Race value="Prot"/>
        <ScoreResult value="BuildOrder"/>
        <EffectArray Reference="Unit,AP_Supplicant,ShieldsMax" Value="25"/>
        <EffectArray Reference="Unit,AP_Supplicant,ShieldsStart" Value="25"/>
        <EditorCategories value="Race:Protoss,UpgradeType:Talents"/>
        <AffectedUnitArray value="AP_Supplicant"/>
    </CUpgrade>
    ```
I'd add instead an icon for the +shield like it was added to Adept and keep +2 shield armor and +25 shield capacity as distinct traits

MadiMadsen:
I disagree for the sole purpose that this would punish the supplicants when the logic for all the other units is 1 upgrade to reach max war council power. All other units only need 1 upgrade.

PR #75, opened 2024-01-22
Free Cloak abilities are replaced with Permacloak Behaviors

EnvyDragon:
For the following units:
- Ghost
- Spectre
- Banshee
- Dusk Wings
- Infested Banshee
- Wraith
- Winged Nightmares

Getting their respective "cloak costs 0 energy" upgrades will now remove the cloak toggle entirely, leaving them permanently cloaked instead.


Ziktofel:
I think that needs generally some discussion as you sometimes let have an uncloaked Ghost in order to take damage instead of Marines

EnvyDragon:
> I think that needs generally some discussion as you sometimes let have an uncloaked Ghost in order to take damage instead of Marines

There was some discussion, and it was very brief; it was broadly decided that the (very niche, but still sometimes present) benefits of toggling cloak were outweighed by the fact that it's much easier to _accidentally_ uncloak a bunch of units than it is to manually recloak just the ones where cloak is free afterward. On top of that, Banshees/Ghosts/Spectres all have non-cloak energy abilities, and if those abilities reduce them to 0 then they uncloak even though their energy isn't being drained. There are way more situations where a free-cloak unit can mistakenly uncloak than where a player would want to uncloak their permacloaked units.

And if a player wants to use a 150/150 Ghost with Crius Suit to tank damage for 50 mineral Marines then something has gone terribly wrong for them.

Ziktofel:
Ghost might be 25/75 and Banshee might have 240 HP

Ziktofel:
Snímek obrazovky pořízený 2024-01-25 20-52-07
The 
Snímek obrazovky pořízený 2024-01-25 20-51-56
Thing is that Crius Suit and Nyx-class cloaking module could go into Cloak's icon spot (instead of using behavior to indicate)

EnvyDragon:
> Ghost might be 25/75 and Banshee might have 240 HP

And you _still_ wouldn't want them to tank for marines, because marines build faster, are cheaper per HP in all circumstances (assuming gas is more valuable than minerals, which is generally the assumption), and have lower DPS than Banshees while lacking the spells that Ghosts/Spectres have. If you're in small enough fights that that sort of tanking is an option, there are probably better uses for that cloak ability. If you're in fights where their ability to tank for marines is worth the risk of losing them, then it probably would have been better to just build more marines.

And again, the handful of grey areas where you _might_ actually consider wanting to tank with a cloaking unit are outweighed by the needless frustration of having to individually cloak them every single time you build them, every time you use their energy on other abilities, and every time you accidentally uncloak them by hitting D.

> screenshots

This would be very good though, I agree. I'll figure out how to implement it and add it to the PR.

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/UnitData.xml
```diff
@@ -2253,8 +2256,30 @@
...
+            <LayoutButtons index="13" Face="AP_GhostSCBWLockdown" Type="AbilCmd" AbilCmd="AP_GhostSCBWLockdown,Execute" Row="1" Column="1"/>
+            <LayoutButtons index="14">
+                <Face value="AP_CriusSuit"/>
+                <Type value="Passive"/>
+                <AbilCmd value=""/>
+                <Requirements value="AP_PermanentCloakGhostResearched"/>
+                <Row value="2"/>
+                <Column value="2"/>
+            </LayoutButtons>
+        </CardLayouts>
```
Ziktofel, line 2282: There shouldn't be any button override. The Crius Suit button belongs to base `AP_Ghost`

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/UnitData.xml
```diff
@@ -2442,8 +2468,28 @@
...
+            <LayoutButtons index="12" Face="Cancel" Type="AbilCmd" AbilCmd="AP_SpectreNuke,Cancel" Row="2" Column="4"/>
+            <LayoutButtons index="13">
+                <Face value="AP_NyxClassCloakingModule"/>
+                <Type value="Passive"/>
+                <AbilCmd value=""/>
+                <Requirements value="AP_PermanentCloakSpectreResearched"/>
+                <Row value="2"/>
+                <Column value="2"/>
+            </LayoutButtons>
+        </CardLayouts>
```
Ziktofel, line 2492: There shouldn't be any button override (all happens on base `AP_Spectre`)

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/UnitData.xml
```diff
@@ -2442,8 +2468,28 @@
     </CUnit>
     <CUnit id="AP_SpectreResourceEfficiency" parent="AP_Spectre">
         <AbilArray index="10" removed="1"/>
-        <BehaviorArray index="2" removed="1"/>
+        <BehaviorArray index="2" Link="AP_SpectrePermanentCloak"/>
+        <BehaviorArray index="3" removed="1"/>
```
Ziktofel, line 2472: Try just cleaner `<BehaviorArray index="2" removed="1"/>`. If it doesn't work, rearrange the behaviors in `AP_Spectre` that the morph is the last

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/UnitData.xml
```diff
@@ -4722,8 +4769,9 @@
         <LifeRegenRate value="0.2734"/>
         <LifeArmorName value="Unit/LifeArmorName/ZergAirArmor"/>
         <ShieldArmorName value="Unit/ShieldArmorName/ZergWildMutation"/>
-        <BehaviorArray Link="AP_RegeneratorRegenBioSteelL1L2"/>
+        <BehaviorArray index="2" Link="AP_RegeneratorRegenBioSteelL1L2"/>
         <BehaviorArray Link="AP_RegeneratorRegenBioSteelL3"/>
+        <BehaviorArray Link="AP_BansheePermanentCloak"/>
```
Ziktofel, line 4774: you don't have to do any changes with Behaviors there

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/UnitData.xml
```diff
@@ -2253,8 +2256,30 @@
     </CUnit>
     <CUnit id="AP_GhostResourceEfficiency" parent="AP_Ghost">
         <AbilArray index="10" removed="1"/>
-        <BehaviorArray index="2" removed="1"/>
+        <BehaviorArray index="2" Link="AP_GhostPermanentCloak"/>
+        <BehaviorArray index="3" removed="1"/>
         <CargoSize value="1"/>
```
Ziktofel, line 2261: See the same with Spectre.

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/UnitData.xml
```diff
@@ -2194,23 +2196,24 @@
         <AbilArray Link="AP_GhostSCBWLockdown"/>
         <AbilArray Link="AP_GhostResourceEfficiencyMorph"/>
         <BehaviorArray Link="AP_OcularImplantsIndicator"/>
-        <BehaviorArray Link="AP_CriusSuitIndicator"/>
         <BehaviorArray Link="AP_GhostResourceEfficiencyMorph"/>
+        <BehaviorArray Link="AP_GhostPermanentCloak"/>
```
Ziktofel, line 2200: Why not to just replace the row with the other stuff. Then you don't have to touch the Resource Efficiency variant

Ziktofel (changes requested):
Do some cleanup

PR #80, opened 2024-01-28
sc2: Moved protoss unit production to submenus rather than behaviours

MatthewMarinets:
Reworked the Protoss build command card locally to use the submenu functionality rather than using behaviours to show and hide buttons.

## Sticky keys
* Buildings have a base submenu with multiple faction specific submenus
* When you build a unit through a submenu, that unit then takes over its slot in the base submenu
* The base submenu button selection is shared across all buildings of that type, including across warp and non-warp variants
* Default submenu state on startup is based on the button priority in UnitData.xml
* Demo: https://youtu.be/rZ_HagE9zMo

## Known issues
* Currently exit submenu and cancel unit are on the same hotkey and in the same positioning; the user may intend to cancel and may just go back to the main menu without cancelling
* Hotkey menu only allows for 4 layouts; the Tal'Darim layout gets cut off
  * Currently not much of an issue as all the button hotkeys are linked for the slot
  * If we unlink the hotkeys, we can make dummy buildings with the tal'darim hotkey layout specifically for hotkey rebinds
* Complexity got moved from game data files to triggers; tables in the trigger must be kept up-to-date with button positions in UnitData.xml and train ability indices in AbilData.xml
  * Could work around this with some code generation if it's a big issue

## Sticky buttons Implementation notes
* The buttons are filtered by activating / deactivating abilities through `TechTreeAbilityAllow` in a trigger
* To filter base menu without affecting submenus, those are different abilities. Base menu uses `xxxTrainRedirect`, which just inherits from `xxxTrain` without modification
* Triggers are listening for UnitOrder events on Protoss train / warpTrain abilities, and call a Galaxy function
* The function in Galaxy script takes a building index and a train ability index and uses lookup tables to get a list of train ability indices for that button slot. It disables all abilities in that slot except the one corresponding to the ability that was used

## Testing
This command gives all protoss units and warp harmonization:
`?GiveProtossTech 1073741823 511 0 0 0 0 16`

Use `?GiveResources <minerals amount> <gas amount> <supply amount>` to give yourself money for testing.


MatthewMarinets:
Posting here that this PR is ready for review / merge; not sure if that message got lost amid other messages.

Ziktofel:
Bug: Warp Harmonization isn't working porperly:
[cinnamon-2024-02-01T145411+0100.webm](https://github.com/Ziktofel/Archipelago-SC2-data/assets/3447808/f3c6820e-ac47-49a2-a477-95e25e7a8a45)

Ziktofel:
BTW it'd be IMHO better if the submenus weren't sticky (return to main command card and update the button)

MatthewMarinets:
Pushed some quick changes; looks like warping in supplicants is broken somehow. Won't have time to look more into it until late this evening.

Ziktofel:
Warp Train of Supplicants is actually some kind of a magic trigger

MatthewMarinets:
Updated the trigger, Supplicant warp-in worked properly from the redirect ability. Everything should be good to go afaik

MatthewMarinets:
Found another issue: receiving items resets the default submenu's unit selection. I'd be happy to merge this and fix that in another PR; can look into it in the next few days otherwise.

Ziktofel:
I don't want to introduce a game-breaking bug during the beta async. Now I'm busy with other issues

PR #85, opened 2024-01-31
Fix Kerrigan bonus levels

Salzkorn:
Since it's been reported as a bug on several occasions, this changes the bonus levels for Supreme/The Infinite Cycle to be a minimum rather than an additive bonus, to be more in line with other granted tech.

Tested by opening Supreme with a level 0 Kerrigan and a level 70 Kerrigan.


MatthewMarinets:
This is one of the main things keeping me banning no-builds with Kerrigan in them, really hope it makes it in.

Ziktofel:
I think in order to avoid any reports we have to drop this feature at all (and make a logic rule for kerrigan levels, if not possible, return always false and fail to generate if this mission occurs)

Ziktofel:
Due to: https://github.com/Ziktofel/Archipelago/pull/152 it should be for Grant Story Tech only. Also update the variable names for better clarity

Ziktofel:
Redesigned and replaced with something else

PR #113, opened 2024-04-02
sc2: Removed two unused upgrades from UpgradeData.xml

MatthewMarinets:
Removed some unused upgrades from UpgradeData.xml. I noticed them when checking to make sure all BC upgrades also affect Jackson's Revenge, and these ones didn't fit that criterion.

Tested that I was still able to start up the game fine, but didn't do any testing beyond that -- this seems to be dead code so I don't see how it could affect anything.

## Details -- removed:
* Upgrade AP_EnergyStorage
* Upgrade AP_BattlecruiserBehemothReactor
  * RequirementNode AP_CountUpgradeBattlecruiserBehemothReactorQueuedOrBetter
  * RequirementNode AP_NotCountUpgradeBattlecruiserBehemothReactorQueuedOrBetter
  * Requirement AP_LearnBattlecruiserEnergyUpgrade


Ziktofel:
Yeah, that's some stuff from melee. I think the Behemoth Reactor could get back later with a different implementation (-25 energy for all BC spells as in HotS/LotV campaign yamato costs only 100 energy)

MatthewMarinets:
In that case, what exactly should be kept? The way the upgrade is incorrect right now (doesn't affect Merc BCs), and the requirement nodes / validators on it wouldn't be necessary for the type of upgrade it is. Should I bring just the upgrade back and update it? I myself am in favour of removing it until we actually want to add it properly.

PR #119, opened 2024-04-04
Refactor Guardian Shell into a Player Response that only triggers when SoA passives are active

MindHawk:
First time contribution so let me know if I made any mistakes

I did not remove the behavior from all the units that have it to easily restore previous functionality if this has some unforeseen bug. This should be done later if there are no issues.


Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/PlayerResponseData.xml
MatthewMarinets, line 1: Do you know a good source of documentation for all the different datatypes somewhere? I would have no idea how to discover something like this on my own.

MindHawk: It's mostly experience and asking the right people, but sc2mapster and its associated discord is usually a decent source.

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/PlayerResponseData.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Catalog>
+    <CPlayerResponseUnitDamage id="SoAGuardianShell">
```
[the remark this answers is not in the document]
MindHawk: Fixed, thanks!

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/PlayerResponseData.xml
MatthewMarinets: I've poked around in there a little bit when trying something new, I'm just used to Python documentation and wish to see it in more places. Will keep looking there in the future, then, and see if I might be able to share something in return someday.

MatthewMarinets:
I don't see any obvious switch within the triggers controlling whether this behaviour applies to terran / zerg. We have a yaml option controlling whether SOA abilities apply in terran / zerg missions and another controlling whether it applies to no-builds. Currently these options are passed to the map on startup with the `?SetOptions` chat message. Should that be handled in the triggers here somehow, or are you expecting to add logic on the client-side to withhold the upgrade depending on mission and options? (or did I just miss this in the triggers?)

MindHawk:
It is handled by the triggers because it checks for the existing of the autonomous SoA caster

MatthewMarinets (approved):

Ziktofel:
Requires generator change in order to handle these groups properly

MindHawk:
I don't understand why it requires a generator change, could you elaborate?

Ziktofel:
if you have a case that SoA autocasts are never usable in your game, you can get Guardian Shell item now, that'd do nothing

MatthewMarinets:
> if you have a case that SoA autocasts are never usable in your game, you can get Guardian Shell item now, that'd do nothing

I think this just means we get our first real branch between sc2-next and master? So long as this change doesn't make it to the `/download_data` package outside of the next beta client, it should be safe to merge.

Ziktofel:
I want it to merge after making the required client part

PR #120, opened 2024-04-04
Adding core faction-swap methods to AP_Player mod

EnvyDragon:
Adding methods to the framework to apply faction-swapping to maps that are set up for it.

How it works:
- Takes one unit group at a time, and a target faction to convert to (most of the time it'll be the player's starting units, but you CAN just make multiple groups if you want to convert multiple players to different factions)
- Runs through every unit in the group, converts refinery structures, converts workers, and converts Town Hall structures
- If structure is a Hatch/Lair/Hive, larva surrounding it will be removed first, then structure will be converted, then surrounding creep will be removed
- If target faction is Zerg, CC/Nexus is converted to a Lair, all larva are removed to normalize count, then creep is added if necessary and AP_Larva are spawned
- Everything else that doesn't belong to the target faction gets removed (and creep tumors have surrounding creep removed as well)
- This means devs WILL need to add other starting structures/units on a per-map basis, and WILL need to consider things like cutscenes, or missions that give the player a base later than usual.

How to apply it to a map:
- Add whatever starting units you want for the other factions (not including workers, gas, or town hall structures).
- NOTE: If you want to know how far creep would extend from a potential Lair, you _can_ just stack a lair on top of a CC/Nexus; the triggers can handle this situation.
- Select all the player's starting units **EXCEPT HEROES** and add them to a unit group called something like `Starting Units."
- Go into the map's triggers and find ArchInit. Add "Remove Unused Starting Units" as a trigger and fill in the parameters as necessary.


Ziktofel:
So if the building is not in those groups, it won't be replaced? (Missions where the player has 2 races)

And thinking about maps Lab Rat, where the player is Terran (but has all Zerg units) and The Essence of Eternity (where the player starts as Zerg and after the initial stage is Terran. Will this method cause problems?

EnvyDragon:
Something like 3/4 of maps across all campaigns will need some kind of special treatment, whether it's for rescuable units, heroes, map objectives, etc, and these core methods don't handle any of that. All these methods do is facilitate the replacement of base structures, and it's all opt-in. Nothing happens to a map until someone deliberately adds these actions to the map's triggers.

For Essence of Eternity, a dev could put the Remove Unused Starting Units trigger after the swap to Terran; for Lab Rat, nearly everything about that mission will need some form of special handling.

MatthewMarinets:
Rather than getting the player race from the `PlayerRace(lp_player)`, why not just pass it in? I don't know what the intended behaviour is for controlling AI allies, but I feel like it may be more helpful to specify both the race being transitioned from and being transitioned to for cases like that where one player may have more than one race's units under their control.

Ziktofel:
I'll link what I've written to Discord:
https://discord.com/channels/731205301247803413/1101186175722598521/1225588672116690955
I think we should use the pattern like difficulties have (onDifficultyCasual, etc triggers)
thus we'd have onRaceSwapTerran, etc.
the worker/refinery/main building replacement would be a function accepting a unit group
(and a race to replace to)
instead of hard-wiring to the player's set race
the onDifficulty pattern aloows us to do more htan just remove units from a given unit group and some missions use that

EnvyDragon:
Agreed on all your comments by the way, I just haven't gotten around to refactoring. I'll come back to it soon.

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/Lib15EF4C78.galaxy
```diff
@@ -2874,6 +2874,190 @@ int lib15EF4C78_gf_AP_Player_getKerriganLevel (int lp_player) {
...
+                UnitSetPropertyFixed(libNtve_gf_LastReplacedUnit(), c_unitPropResources, lv_initialGas);
+            }
+
+        }
+        else if (auto148E77B6_val == "Prot") {
+            if ((UnitGetType(lp_gasStructure) != "AP_Assimilator") && (UnitGetType(lp_gasStructure) != "AP_AssimilatorProtossGeyser") && (UnitGetType(lp_gasStructure) != "AP_AssimilatorPurifierGeyser") && (UnitGetType(lp_gasStructure) != "AP_AssimilatorShakurasGeyser")) {
+                libNtve_gf_ReplaceUnit(lp_gasStructure, "AP_Assimilator", libNtve_ge_ReplaceUnitOptions_NewUnitsDefault);
+                UnitSetPropertyFixed(libNtve_gf_LastReplacedUnit(), c_unitPropResources, lv_initialGas);
+            }
+
```
Ziktofel, line 2952: I'd also check against the automated/rich variants in order to avoid race conditions/possible bugs

EnvyDragon: Given that rich variants don't exist in the base game, I didn't handle them at all, particularly because they'd need to be handled differently. 

Agreed on autogas though, I've added those in.

PR #126, opened 2024-04-14
Fixing no-build skip-cutscene bugs

EnvyDragon:
Belly of the Beast:
- Fixing a bunch of bugs related to the skip-cutscene flag; timer never being removed etc (this was VERY early on in the cutscene-skip process, the flag just wasn't applied properly)

Back in the Saddle:
- Fixing a bug where the Umojan forces wouldn't be removed before the Archangel fight if skip-cutscenes was active, resulting in allied siege tanks helping in the boss fight


Ziktofel:
cherry-picked to master

PR #127, opened 2024-04-16
Adding hard/brutal larva behavior

EnvyDragon:
In vanilla HotS, hatch/lair/hive spawns larva:
- Once every 9 seconds on Casual/Normal, once every 12 seconds on Hard/Brutal
- Up to a cap of 9 on Casual/Normal, up to a cap of 6 on Hard/Brutal

Previously AP only had the casual/normal variation; this PR adds the Hard/Brutal version.

Effect is applied as an upgrade; upgrade was formerly applied in "give zerg default tech" trigger, but was moved to "set difficulty" trigger to avoid race conditions.


MatthewMarinets (approved):
Just looked through the changes, didn't test anything. Style looks good and content all makes sense.

Ziktofel:
Looks like I've added a TODO after implementing that thing first

PR #130, opened 2024-05-12
sc2: adding detection circles

MatthewMarinets:
Targeting master because this is a bugfix that can affect the live maps.

Trying to add detection circles back in for AP units.

Current problem is that both vanilla and my detection circles can render at the same time:
Screenshot2024-05-11 21_32_38

Making a draft PR to seek help on that issue.


MatthewMarinets:
Technically vanilla NCO has a different graphic for detection circles, but I couldn't find where it was defined exactly. We could override it in ArchipelagoPlayerNCO, or I'd even be in favour of making it the default because it looks rad.

MatthewMarinets:
Found the NCO graphics. Old pictures with lotv graphics for comparison:

Screenshot2024-05-11 22_07_19
Screenshot2024-05-11 22_13_51
Screenshot2024-05-11 23_35_16

Ziktofel:
I've fixed some units not working for this feature

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/Lib15EF4C78.galaxy
```diff
@@ -247,6 +247,26 @@ void lib15EF4C78_InitCustomScript () {
...
+
+    // Variable Initialization
+
+    // Implementation
+    lv_player_index = 2;
+    for ( ; ( (autoAF3B14EE_ai >= 0 && lv_player_index <= autoAF3B14EE_ae) || (autoAF3B14EE_ai < 0 && lv_player_index >= autoAF3B14EE_ae) ) ; lv_player_index += autoAF3B14EE_ai ) {
+        if (((libNtve_gf_PlayerIsEnemy(1, lv_player_index, libNtve_ge_PlayerRelation_Enemy) == true) || (libNtve_gf_PlayerIsEnemy(1, lv_player_index, libNtve_ge_PlayerRelation_EnemyMutual) == true))) {
+            libNtve_gf_SetUpgradeLevelForPlayer(lv_player_index, "AP_DetectionEnemy", 1);
+        }
+
```
Ziktofel, line 266: Unhardcode P1 from there. Instead use check if the controller is User:
Snímek obrazovky pořízený 2024-05-16 15-40-04
Snímek obrazovky pořízený 2024-05-16 15-41-08

Ziktofel (changes requested):

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/Lib15EF4C78.galaxy
MatthewMarinets: I verified some time ago that the human player is player 1 in every campaign mission. If I'm reading the description of this function right, this would involve looping over every player and checking if it's human-controlled. Is it worth the extra visual clutter / for loop for it? Does AP Core have an AP function for this yet?

MatthewMarinets: I don't really have a lot of time today, so I just split the human player ID to be gotten from a function (which for now always returns 1). I figure if triggers always draw from there, we can update the function later if we someday add a map where the human player isn't player 1. If you consider it better to get that done in this PR, I can maybe look into it tomorrow or sometime on the weekend.

image
image

MatthewMarinets:
Squashed and force-pushed so that I could get rid of the binary files from the changed files list. Now I know the verbose variable / function names are changed with ctrl+D I'll go back to assuming DocumentHeader and Triggers.version aren't necessary.

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/Lib15EF4C78.galaxy
Ziktofel: I think that better job is to check if the player is human. This mod's scope is just library of units/upgrades and the human player is always unhardcoded (Vanilla scripts are using P1 but in this mod it's switched into a parameter). This mod is written in a way that can still used for example as a dependency for melee-like map (one of the reasons why it's split from triggers mod).  
What you actually did is to hardcode the P1 into another location.

I'd just loop over the players and check the controller (with the function) and if it's the user, do the rest

MatthewMarinets: Done
image

Did a quick test on Amon's Reach and the detection circles still showed up.

PR #132, opened 2024-05-15
Minor Kerriganless rebalance

EnvyDragon:
On all of the below maps, the following changes were made:
- "Kerriganless Bonus" unit group is hidden by default so that the units don't briefly appear on map load.
- On Kerriganless, Kerrigan gets an extra behavior that suppresses her ability to attack, as security against cases where she inadvertently becomes unpaused. The Kerriganless Bonus unit group also becomes unhidden.

Planetfall:
- Added 1 Brood Queen to the Kerriganless Bonus unit group.

Domination:
- Added 1 Roach to the Kerriganless Bonus unit group.

Rendezvous:
- Added 1 Spine Crawler to the Kerriganless Bonus unit group.

Harvest of Screams:
- Added 1 Spine Crawler to the Kerriganless Bonus unit group (initially owned by Player 2 in the rescuable base).


Magnemania:
Planetfall:  Maybe add a second Brood Queen? Brood Queens don't have Ensnare by default (not that it would be particular helpful against Terran), and Spawn Broodling on a single Marauder or Hellbat isn't very impressive.  If they started with full energy, the player might have something to do at the start of the map (replacing the expected Primal Kerrigan rampage that would occur in vanilla).

Domination: Good, the roach makes it pretty straightforward to get the expansion, helps the pace.

Rendezvous: Good, feels much closer to vanilla now on Kerriganless rather than a frantic rush to get defenses up.  Handles the first two waves but not the third, which is a balanced buffer time.  I wonder if we should consider matching this sort of difficulty in Kerrigan Present by requiring Basic Kerrigan?  Not necessary for this PR, though.

Harvest of Screams: The ice effect isn't applied to the new Spine Crawler:
image
Apart from that, the new Spine Crawler itself is good. Kerriganless can now defend the first post-freeze attack wave without rushing military/cheesing (but takes losses for the greed), so this solves a problem.

Magnemania (changes requested):
Planetfall:  Maybe add a second Brood Queen? Brood Queens don't have Ensnare by default (not that it would be particular helpful against Terran), and Spawn Broodling on a single Marauder or Hellbat isn't very impressive.  If they started with full energy, the player might have something to do at the start of the map (replacing the expected Primal Kerrigan rampage that would occur in vanilla).

Domination: Good, the roach makes it pretty straightforward to get the expansion, helps the pace.

Rendezvous: Good, feels much closer to vanilla now on Kerriganless rather than a frantic rush to get defenses up.  Handles the first two waves but not the third, which is a balanced buffer time.  I wonder if we should consider matching this sort of difficulty in Kerrigan Present by requiring Basic Kerrigan on the logic side?  Not necessary for this PR, though.

Harvest of Screams: The ice effect isn't applied to the new Spine Crawler:
image
Apart from that, the new Spine Crawler itself is good. Kerriganless can now defend the first post-freeze attack wave without rushing military/cheesing (but takes losses for the greed), so this solves a problem.

EnvyDragon:
Fixed the bug on the Spine Crawler on Harvest, and the Brood Queen on Planetfall should now start with full energy. Thanks again for testing!

Magnemania:
> Fixed the bug on the Spine Crawler on Harvest, and the Brood Queen on Planetfall should now start with full energy. Thanks again for testing!

Excellent!  Should be good to go, then.

PR #137, opened 2024-05-24
Adding Objective Categorization - Mar Sara Set

Gemster312:
### Missions Included:
-Liberation Day
-The Outlaws

_Note: Zero Hour is already complete ([PR 131](https://github.com/Ziktofel/Archipelago-SC2-data/pull/131))._

### Bug Fixes / Map Features:
Liberation Day:
> - Made the "Special Delivery" objective appear a bit later into the map instead of at-spawn to reduce confusion
> - Created an objective for the "Transport" location (extra)

The Outlaws:
> - None

### Tests:
- Parital Clear (e.g. finish 50% of an objective, then reload and verify progress is noted)
- All locations enabled, full clear
- All locations enabled, post-full clear re-clear
- Only Vanilla locations enabled
- Only Vanilla and Extra locations enabled
- Only Challenge locations enabled


Maps/ArchipelagoCampaign/WoL/ap_liberation_day.SC2Map/MapScript.galaxy
```diff
@@ -670,25 +682,151 @@ void gt_SixthBonusObjectiveDestroyed_Init () {
...
+    if (!runActions) {
+        return true;
+    }
+
+    TriggerEnable(TriggerGetCurrent(), false);
+    TriggerExecute(gt_ObjectiveTransportCreate, true, false);
+    lib5BD4895D_gf_AP_Core_setObjectiveChecks(gv_objectiveTransport, 0);
+    lib5BD4895D_gf_AP_Core_setObjectiveCompleted(gv_objectiveTransport);
+    lib5BD4895D_gf_AP_Core_sendLocationLocal(gv_LOCATION_ID_TRANSPORT);
+    return true;
```
MatthewMarinets, line 822: On reading this again, I wonder if these sorts of checks (`isLocationCollectable(gv_LOCATION_ID_TRANSPORT) == true`) are better handled on the mod-side. ie, the maps simply specify and objective, what locations it's linked to, and whether it should display even if it has no checks, then the mod can handle all the details on whether to display it / how many checks to display.

I just feel like you're going to have to add over 600 of these if checks, and if we continue down this route there still won't be any connection in AP core mod between location index and objective struct.

MatthewMarinets: I'm imagining something like this:

## State / configuration
AP core maintains a list of objectives (configured by the map), a list of locations that can be collected (supplied by the client on startup). Objectives contain data for:
* What locations they count towards
* Whether they show if there are no locations (ie in-mission progression vs things that are AP-location-only)
* Normal stuff like counter/timer/stuff we already configure

## Internal API
* `AP_Core_locationToObjective(int locationId) -> objective`
* `AP_Core_getObjectiveLocations(objective objective) -> int[]`
* `AP_Core_updateAllObjectiveChecksCounters() -> void`
* `AP_Core_signalLocationId(int locationId) -> void`
  * Sends the location received signal to the client if the location is collectable
  * Updates the `[X checks remain]` counter on the associated objective if necessary
  * I'm not sure if this should handle counters as well or if counters should be handled with a separate function called right next to this function in the map script if necessary

## Usage in map file
The objective categorization largely doesn't change, except some extra information must be passed to objective constructors:
* `int[]` / `int` (depending on constructor) Associated location IDs / a list of associated location IDs
* `bool` Whether this objective gets hidden if it's uncollectable

Whenever a map wants to update an objective or send a location, we update that into a call to `signalLocationId`, which should generally handle everything. If we keep handling counters separately, we will also update the counters with another function then.

MatthewMarinets: I'll note there's another case I'm not sure your test cases cover: what happens when a location is received without reaching that point in the mission? ie someone can `/sendLocation` or `/send` in the console, or someone else in a multiworld can `!collect`. If we spread all the check counter handling out in all the map scripts, we can't easily make updates to handle interactions like this. By providing all necessary data to the mod and handling things there, we only have to make such updates in one place.

There's also some path-dependency going on. If I start a mission with challenge locations enabled, then get sent stuff / find locations, I expect objectives to get crossed off. But if I start the mission with them disabled / everything collected, then I'd expect the objective to not show (assuming enablement and collection state are signaled through the same locations array).

PR #140, opened 2024-05-27
Adding Garrison tip to Enemy Intelligence

EnvyDragon:
Adds a popup tip, intended to perform two functions:
1) Inform new AP players that the mechanic has been changed from base NCO
2) Give players a way to check which units can be garrisoned without needing to build a copy of everything

image

image


EnvyDragon:
Cherry-picking commit so it can go onto master

PR #141, opened 2024-05-27
Garrison Strength tip for Enemy Intelligence

EnvyDragon:
Adding tip to Enemy Intelligence so that:
1) New players are made aware that there's a different mechanic from vanilla
2) All players can see in advance which units can be garrisoned, without needing to build one of everything

image

image


MatthewMarinets (approved):
Code looks good to me at a quick glance.

PR #144, opened 2024-06-16
sc2: Added liberator guerilla missiles to replace 360 degree sensors

MatthewMarinets:
Replaced liberator 360 degree sensors upgrade with Guerilla Missiles upgrade:

Liberators in Fighter Mode apply a debuff to enemy units the hit, causing them to move and attack slower for 8 seconds.

I'm fine to play with the numbers a little bit. Currently it's 30% reduced move speed and 15% reduced attack speed (see `AP_LiberatorGuerillaMissilesDebuff` in BehaviorData.xml). The move speed is much stronger, but takes kiting to really utilize and that's hard to do with the liberators' slow damage point and short range; it also doesn't synergize with sieged liberators as much. The attack speed debuff is much more subtle, but provides some reason to mix in a few liberators with an AA mix to reduce incoming damage.

The visual effect could also be adjusted on request, but I'm going to leave it as-is until someone provides feedback / says how they'd like it changed. Currently it uses a scaled-down version of Fenix's Whirlwind effect animation.

image

image

video link:
https://discord.com/channels/731205301247803413/980554570075873300/1251799672234709032


MatthewMarinets:
Got it to grey out in def
image

MatthewMarinets:
image

PR #174, opened 2024-07-08
sc2: replaced skylord interceptors with a laser

MatthewMarinets:
Reworking skylord to be different from a reskinned carrier. Gave it a slow but high-damage laser that deals damage in a line.

WIP
- [x] Apply attack upgrades
- [x] Test Squad Sight
- [ ] ~~Change teleport around -- maybe shorter cooldown / activation, but shorter range?~~ Going to leave this for another day for time/availability reasons
- [x] Fix a bug where one attack can hit the same unit multiple times over

### DIFF
--- a/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/ActorData.xml
+++ b/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/ActorData.xml
@@ -32874,9 +32874,9 @@
         <On Terms="ActorCreation" Send="SetTintColor 255,70,70"/>
         <On Terms="Effect.AP_CarrierTaldarimLaserDamage.Start; At Caster; FromEffectTreeDescendant" Send="ActionDamage"/>
         <On Terms="Effect.AP_CarrierTaldarimLaserCP.Start; At Caster" Send="Create"/>
+        <On Terms="Effect.AP_CarrierTaldarimLaserCPSquadsight.Start; At Caster" Send="Create"/>
         <LaunchAssets Model="" Sound="AP_SiegeBreaker_SiegedAttackLaunch"/>
         <BeamScope value="Caster"/>
-        <Missile value="HellionAttackBaseMissile"/>
         <ShieldFlashType value="Header"/>
     </CActorAction>
     <CActorBeamSimple id="AP_CarrierTaldarimLaserAttackBeam" parent="GenericAttackBeam">
--- a/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/ActorData.xml
+++ b/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/ActorData.xml
@@ -32859,6 +32859,44 @@
     <CActorModel id="AP_CarrierHyperjumpTeleportOut" parent="ModelAnimationStyleOneShotNoOrphan">
         <Model value="AP_AirUnits_TeleportOut_COOP"/>
     </CActorModel>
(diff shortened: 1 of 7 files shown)


MatthewMarinets:
image
skylord_test
image
image
image

+tested that buffing range with havoc leads to increased range and still does damage at the end of the range.

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/ActorData.xml
```diff
@@ -32859,6 +32859,43 @@
...
+        <On Terms="ActorCreation" Send="SetScale 1.000000"/>
+        <On Terms="TimerExpired; TimerName DelayDestroy" Send="Destroy"/>
+        <On Terms="ActorCreation" Send="TimerSet 0.300000 DelayShrink"/>
+        <On Terms="TimerExpired; TimerName DelayShrink" Send="SetScale 0.400000"/>
+        <HostSiteOps Ops="SOpShadow"/>
+    </CActorModel>
+    <CActorAction id="AP_CarrierTaldarimLaserAttack" parent="GenericAttackBase">
+        <On Terms="ActorCreation" Send="SetTintColor 255,70,70"/>
+        <On Terms="Effect.AP_CarrierTaldarimLaserDamage.Start; At Caster; FromEffectTreeDescendant" Send="ActionDamage"/>
+        <On Terms="Effect.AP_CarrierTaldarimLaserCP.Start; At Caster" Send="Create"/>
```
Ziktofel, line 32876: Missing `AP_CarrierTaldarimLaserCPSquadsight` (the attack visuals and sound won't be played if under Havoc's buff)

Ziktofel (changes requested):
Bug: You can't attack own units if under Havoc's range buff (you can without that)

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/ActorData.xml
MatthewMarinets: This is a quick fix, commit pending on the other issue.

MatthewMarinets:
> Bug: You can't attack own units if under Havoc's range buff (you can without that)

I can't seem to find a place to fix this. Would you happen to know what causes this behaviour? My best guess is that _specifically for squadsight_ the aoe damage triggers first, which markers the friendly unit but doesn't deal damage because of the filters, then the single-target damage doesn't trigger because of the marker. Not sure if that's the case or how to get around that, though?

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/EffectData.xml
```diff
@@ -7772,6 +7772,57 @@
...
+        <EditorCategories value="Race:Protoss"/>
+        <Kind value="Ranged"/>
+        <KindSplash value="Splash"/>
+        <Amount value="55"/>
+        <Death value="Fire"/>
+    </CEffectDamage>
+    <CEffectEnumArea id="AP_CarrierTaldarimLaserDamageArea">
+        <EditorCategories value="Race:Protoss"/>
+        <ExcludeArray Value="Target"/>
+        <IncludeArray Effect="AP_CarrierTaldarimLaserCP" Value="Target"/>
```
EnvyDragon, line 7786: Not certain about this; should this be AP_CarrierTaldarimLaserEffect? (or maybe there should be a second entry?) It looks to me like it's including the target of the regular version, but not the Squadsight version

Issue #152, opened 2024-07-01
Protoss War Council rebalance checklist


MatthewMarinets:
Creating a checklist for the war council rebalance sheet here: https://docs.google.com/spreadsheets/d/1hObgY0_T7O94AOJ6wVCElEex-sE6JKQmIMsQO2oshpA/edit

Some upgrades are already implemented as default tech in the `AP_Triggers_giveProtossDefaultTech()` action, I'll try and capture them here. @MadiMadsen would be nice if you could confirm what you implemented already, I see Zealot, Centurion, Sentinel, Void Ray, Phoenix, etc. I also notice some Supplicant and Slayer upgrades that seem unrelated. I am also unsure what the recommended testing configuration is, I'm hoping to shift things over to a client option soon.

- [x] Zealot
- [x] Centurion
- [x] Sentinel
- [ ] Supplicant
- [ ] Dragoon
- [ ] Stalker
- [ ] Adept
- [ ] Instigator
- [ ] Slayer
- [ ] Avenger
- [ ] Dark Templar
- [ ] Blood Hunter
- [ ] Sentry
- [ ] Energizer
- [ ] Havoc
- [ ] High Templar
- [ ] Signifier
- [ ] Dark Archon
- [ ] Ascendant
- [ ] Immortal
- [ ] Annihilator
- [ ] Vanguard
- [ ] Reaver
- [ ] Colossus
- [ ] Wrathwalker
- [x] Phoenix
- [ ] Scout
- [ ] Corsair
- [ ] Mirage
- [ ] Arbiter
- [x] Void Ray
- [ ] Destroyer
- [ ] Carrier
- [ ] Tempest
- [ ] Mothership

MadiMadsen:
Basically I managed to do the entire zealot line so far, supplicant was the last one I completed.
I used the unit test map and basically tested the unit until I had the proper rebalanced version and then I gave it a research trigger and turned it on and to make sure the upgrade worked properly. 
And for organization sake I put all these upgrades to be on by default in archipelago triggers together I did not worked on anything above the comment so I cannot validate that those are fully functional.

MadiMadsen:
small correction: it looks like the changes i made to supplicant were undone
one of zik's recent commitments undid the upgrades given to supplicants by default for the purposes of this rebalance.

MatthewMarinets:
Noting some feedback and ideas that I won't be able to implement during the first half of holiday time:

### Resource efficiency
[...]

PR #199, opened 2024-07-17
sc2: Added slayer and avenger war council upgrades -- phase blink, kryhas cloak

MatthewMarinets:
Added slayer war council upgrade: Phase Blink.

Without it, they can't blink. With it, they can blink. Next attack within 8 seconds after blinking does double damage.

Pairs with [client PR #237](https://github.com/Ziktofel/Archipelago/pull/237)
Progress on #152 

Without upgrade:
image

With upgrade:
image


MatthewMarinets:
Tossing in Avenger since I have a little more time.

Upgrade: Kryhas Cloak (thanks ShadowArchon for the name suggestion). Gives Avengers their permacloak

Without upgrade:
image

With upgrade:
image

PR #208, opened 2024-07-23
sc2: Dark Templar War Council Upgrades - Shadow Fury

MatthewMarinets:
Added Dark Templar War Council Upgrades -- Shadow Fury.

Shadow Fury is so strong I decided to split it in two. The "Lesser" variant gives 2 strikes, the "Greater" one gives 3.

Progress on #152 
Pairs with [client PR #247](https://github.com/Ziktofel/Archipelago/pull/247)

Without upgrade:
image

With Greater only:
image

With Lesser only:
image

With both:
image

* Tested all cases output the correct number of slashes and reappeared / became vulnerable again after the last slash.
* Tested unlocks work with client `/send` commands

#### TODO
- [x] Add unlock triggers
- [x] Test with client


MatthewMarinets:
Just realized this likely breaks predator shadow fury -__-

I'm in favour of removing predator shadow fury to replace it with a different upgrade. The alternative is to split the relevant parts of the predator ability, which I think just involves splitting 2 effects and a behavior

Ziktofel (changes requested):
```
[7/23/2024 8:34:58 PM] Warning: XML: \Program Files (x86)\StarCraft II\Mods\ArchipelagoPlayer.SC2Mod\Base.SC2Data\GameData\BehaviorData.xml(2877,9) : Unable to write value: -1; reason: Core: invalid object type identification  Source: Mods/ArchipelagoPlayer.SC2Mod  Entry: AP_DarkTemplarShadowFuryController  Scope: CBehaviorBuff  Field: PeriodCount
```
Snímek obrazovky pořízený 2024-07-23 20-35-33

PR #212, opened 2024-07-25
Added a sub menu for infantry, vehicle and ship units to the Merc Compound

WaikinDN:
There is now free space for additional 4 infantry, 6 vehicles and 3 ships


Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/UnitData.xml
```diff
@@ -7249,25 +7249,37 @@
...
-            <LayoutButtons Face="AP_HireMercWraith" Type="AbilCmd" AbilCmd="AP_SummonMercenaries,10" Row="2" Column="1"/>
-            <LayoutButtons Face="AP_HireMercLiberator" Type="AbilCmd" AbilCmd="AP_SummonMercenaries,11" Row="2" Column="2"/>
-            <LayoutButtons Face="AP_HireBrynhildFighter" Type="AbilCmd" AbilCmd="AP_SummonMercenaries,12" Row="2" Column="3"/>
-            <LayoutButtons Face="AP_HireMercThor" Type="AbilCmd" AbilCmd="AP_SummonMercenaries,13" Row="2" Column="4"/>
+            <LayoutButtons Face="AP_MercVehicle" Type="Submenu" SubmenuCardId="0002" Row="2" Column="1"/>
+            <LayoutButtons Face="AP_MercShip" Type="Submenu" SubmenuCardId="0003" Row="2" Column="2"/>
             <LayoutButtons Face="Halt" Type="AbilCmd" AbilCmd="BuildInProgress,Halt" Row="2" Column="3"/>
             <LayoutButtons Face="CancelBuilding" Type="AbilCmd" AbilCmd="BuildInProgress,Cancel" Row="2" Column="4"/>
         </CardLayouts>
+        <CardLayouts CardId="0002">
```
MatthewMarinets, line 7264: Right now, if you select any vehicle or ship merc, the submenu will revert back to the infantry tab. You can set a flag on the submenus ("sticky" I think?) to make it stay on the submenu when you use an ability. It will reset if you deselect+reselect the building. Not sure what behaviour you think would work best.

WaikinDN: My bad. I'll get it fixed

MatthewMarinets: I wouldn't call it a bad, I was just pointing out another direction you could take the design. Either could work.

Ziktofel: Before Phan's changes to Protoss building there was behavior-driven stuff instead using submenus. It worked in a way that it looked like submenus but was sticky across re-selecting.

The implementation of this trick relies on behaviors and having requirements for each group that require behavior on the unit. The tab selection button enables the behavior so the correct options are shown.

I think it could be the desired behavior there as I set Vehicles and it stays so then across re-selecting

WaikinDN: A second commit was done to make it behave as desired.
Now if I call a mercenary from a submenu, the submenu will not change.

MatthewMarinets: re: Behaviour-based things, that was quite a complicated system (the new system is also complicated), and I don't think that level of complexity is necessary here. The fastest default merc cooldowns are 5 minutes, (2m30 with cooldown speed, 4minutes with proposed balance changes in chat), that's slow enough that I think it's reasonable to select the submenu every time you're calling something down.

WaikinDN: The merc compound can be used as an emergency base defense tool. If the submenu defaults to infantry after every calldown, it may disorient some players in stressful situations.

Ziktofel: it should be left on the last visited tab and not default to an empty tab if possible on start

MatthewMarinets: Pretty sure there is no empty tab, it will default to infantry.

Ziktofel:
I've polished it a bit and changed the underground mechanism for the tabs.

Thinking if we could use the charge system to display how many mercs are ready on each tab

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/UnitData.xml
```diff
@@ -7243,42 +7243,64 @@
...
         <BehaviorArray Link="UnderConstruction"/>
+        <BehaviorArray Link="AP_MercCompoundTabNotSelected"/>
         <CardLayouts>
+            <!-- Infantry -->
             <LayoutButtons Face="AP_HireKelmorianMiners" Type="AbilCmd" AbilCmd="AP_SummonMercenaries,Train1" Row="0" Column="0"/>
             <LayoutButtons Face="AP_HireDevilDogs" Type="AbilCmd" AbilCmd="AP_SummonMercenaries,Train2" Row="0" Column="1"/>
             <LayoutButtons Face="AP_HireHammerSecurities" Type="AbilCmd" AbilCmd="AP_SummonMercenaries,Train4" Row="0" Column="2"/>
             <LayoutButtons Face="AP_HireMercReaper" Type="AbilCmd" AbilCmd="AP_SummonMercenaries,Train10" Row="0" Column="3"/>
             <LayoutButtons Face="AP_HireMercMedic" Type="AbilCmd" AbilCmd="AP_SummonMercenaries,Train9" Row="0" Column="4"/>
+            <!-- Vehcicles -->
```
MatthewMarinets, line 7261: Typo in Vehcicles

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
```diff
@@ -7174,4 +7174,22 @@
...
+    </CBehaviorBuff>
+    <CBehaviorBuff id="AP_MercCompoundTabNotSelected">
+        <InfoFlags index="Hidden" value="1"/>
+        <EditorCategories value="AbilityorEffectType:Units"/>
+        <RemoveValidatorArray value="AP_NotMercCompoundInfantryTabSelected"/>
+        <RemoveValidatorArray value="AP_NotMercCompoundVehiclesTabSelected"/>
+        <RemoveValidatorArray value="AP_NotMercCompoundStarshipsTabSelected"/>
+        <Period value="1"/>
+        <PeriodicEffect value="AP_MercCompoundTabSearchSwitch"/>
+    </CBehaviorBuff>
```
MatthewMarinets, line 7420: Considering there's nothing _on_ the misc tab, maybe it's better to default to the infantry tab like Waikin had?

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/EffectData.xml
MatthewMarinets, line 1: Rather than using behaviours, which are per-merc compound, maybe it's better to set / remove upgrades (there should be an effect for that) so that it's player-wide?

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
Ziktofel: The first tab that gets available is opened. So if you have only starships, the starship tab will open on mission start. This basically waits for anything getting available

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/EffectData.xml
Ziktofel: Thinking how it'd work with upgrades:

I think we'd need trigger that listens to a dummy Effect. This triggers sets up the upgrades

MatthewMarinets: Just checked, CEffectModifyPlayer can give / remove upgrade levels. I'm not sure if it can _set_ instead of just add, which would make things easier. But this is possible without triggers.

Ziktofel: Looks like there's no operation. I'll try triggers

WaikinDN:
I see that a lot of changes were made. How easy would be to add more buttons now?
I ask because I'd like to eventually add the "Special Units" button and maybe in the future an "Outlaws" button for Tychus friends if there is a demand for them.

Ziktofel:
A new tab would require 2 empty upgrades - one for the tab itself to enable, the other for storing the current selected tab

Also update of a simple trigger and an abil to switch into that tab

Harder would be the part for giving the player an idea how many mercs ready to go he has on every tab

PR #220, opened 2024-07-27
sc2: Adding Skirmisher and Vanguard War Council upgrades

MatthewMarinets:
Adding Skirmisher war council upgrade: Peer contempt. Nerfed unit can only target ground, war council version can hit air.

I noticed skirmishers felt really weak, though that may just be because I was trying on temple of unification and there is lots of armoured AA there. Considering giving them a range and/or damage buff to compensate for their lack of disable.

Adding a new item category because war council 1 will be mostly filled up just with ground units.

Progress on #152
Pairs with [client PR #256](https://github.com/Ziktofel/Archipelago/pull/256)

## Pictures
Without upgrade:
image
image
image
image

With upgrade:
image
image
image
image


MatthewMarinets:
Throwing Vanguards into the mix. 2 upgrades as their DPS is just that insane:
* Agony Launchers: Gives +7 damage vs armoured (per attack). Comes from the Alarak upgrade.
* Rapidfire Cannons: Makes damage period go from 6 to 3.75. When nerfed, it makes them more like slow-reloading artillery

### Pictures
#### Without upgrades:
image
image
(also gave the weapon the cooldown visual):
image

#### With upgrades:
image
image
image
image

PR #221, opened 2024-07-28
Haven's Fall race-swap

EnvyDragon:
Generator part is here: https://github.com/Ziktofel/Archipelago/pull/259

Adding Zerg and Protoss versions of Haven's Fall.


Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/WeaponData.xml
```diff
@@ -1734,7 +1734,7 @@
         <Icon value="Assets\Textures\btn-upgrade-zerg-airattacks-level0.dds"/>
         <DisplayEffect value="AP_ParasiteSpore_Damage"/>
         <TargetFilters value="Air,Visible;Missile,Stasis,Dead,Hidden,Invulnerable"/>
-        <Period value="1.9"/>
+        <Period value="1.36"/>
```
Ziktofel, line 1737: What's this change?

Ziktofel (changes requested):

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/WeaponData.xml
EnvyDragon: Attack speed buff for the Corruptor so it lines up with the [current multiplayer version](https://liquipedia.net/starcraft2/Corruptor_(Legacy_of_the_Void)). Necessary because otherwise the corruptor is just not very good, and it's the only zerg anti-armored anti-air available for the mission.

Ziktofel: That linked value is for fastest speed the MP displays values for. The real value is still 1.9

EnvyDragon: I'm not sure what you mean? The WoL/HotS versions had 1.9, the current/live one has 1.36

EnvyDragon: Ah, I think I see what you mean -- it was a visual change so that attacks speeds etc would align more closely with "faster" gamespeed. 

I still think corruptors absolutely need a buff, and they feel a lot better at this speed; when attacking Brood Lords, the Viper is practically a better choice despite being a dedicated caster. I don't think it necessarily needs to happen in this PR though, so I'll remove it.

PR #229, opened 2024-08-01
Fix Terran version of Haven's Fall

itsjustbones:
currently, if you load the terran version of haven's fall, you get the zerg version of the map instead.  This is due to an override in player race, presumably for testing purposes.  This change just removes that override.

I tested by loading in to all three variants of the mission after generating the map, all three races worked.


itsjustbones:
@EnvyDragon Might want to double check in case I broke something I don't understand, but I'm pretty sure it's a simple fix.

EnvyDragon:
Yep, that's the debug trigger I left in by accident! Thanks for catching it.

PR #239, opened 2024-08-03
sc2: stalwart war council upgrade

MatthewMarinets:
Adding the Stalwart war council upgrade -- duality charge. Nerfed unit has -4 range (6 -> 2), item adds the +4 range back.

The name is a pun on being twice as good as singularity charge. I placed the passive icon on the bottom row to match the style of immortals.

Progress on #152
Pairs with [client PR #266](https://github.com/Ziktofel/Archipelago/pull/266)

### Without upgrade
image
image

### With upgrade
image
image


MatthewMarinets:
(forgot to push the trigger commit, but it's unchanged from when I tested it)

PR #255, opened 2024-08-08
sc2: Adding Corsair war council upgrade -- network disruption

MatthewMarinets:
Adding corsair war council upgrade -- Network Disruption. Nerfed version has 1/3 radius on dweb, upgrade restores it back.

Tested it a little with 1/3 and 1/2 radius, both felt fine. The auto-cast seems to spread things out a little more with 1/3, and vs e.g. stalkers this can create a sparser field that still doesn't let stalkers fire from the cracks:
corsair_wc_3

Progress on #152
Pairs with [client PR #279](https://github.com/Ziktofel/Archipelago/pull/279)

### Without upgrade
image
image

### With upgrade
image
image


Ziktofel (changes requested):
See generator PR

PR #259, opened 2024-08-11
sc2: Added Void Ray war council upgrade -- Prismatic Range

MatthewMarinets:
Adding Prismatic Range -- nerfed void rays don't gain range as the beam charges. Also doing the actor data stuff to get the melee/oldschool models pre-upgrade.

Progress on #152
Pairs with [client PR #285](https://github.com/Ziktofel/Archipelago/pull/285)

### Without upgrade
image
image
image
image
image

### With upgrade
image
image
image
image
image


MatthewMarinets:
Updating the warp ray wireframe so that it's different from the void ray one, meaning things are actually distinguishable in the command card:

image
image

PR #268, opened 2024-08-16
sc2: Added Observer war council upgrade -- scopophobia inducers

MatthewMarinets:
Added Observer war council upgrade -- Induce Scopophobia. Applies a debuff on enemies that reduces their movement and attack rate.

Progress on #152
Pairs with [client PR #290](https://github.com/Ziktofel/Archipelago/pull/290)

Also giving Adepts higher selection priority than Avengers:
image

### Without upgrade
image

### With upgrade
image
image
observer2
image
image


Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/ActorData.xml
```diff
@@ -14220,6 +14220,15 @@
...
     </CActorUnit>
+    <CActorModel id="AP_Scophophobia" parent="ModelAnimationStyleContinuous">
+        <Inherits index="BaseModelScale" value="1"/>
+        <Inherits index="Scale" value="1"/>
+        <Inherits index="Visibility" value="1"/>
+        <On Terms="Behavior.AP_ScopophobiaTarget.On" Send="Create"/>
+        <On Terms="Behavior.AP_ScopophobiaTarget.Off" Send="AnimBracketStop BSD"/>
+        <Host Subject="_Unit"/>
+        <HostSiteOps Ops="SOpAttachCenter"/>
+    </CActorModel>
```
Ziktofel, line 14231: Range actor for the ability?

Ziktofel (changes requested):

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/ActorData.xml
MatthewMarinets: I don't think it's worth; the range is set to be the same as vision range, the observers will autocast it so it doesn't need conscious attention, the activation is obvious so strategy is straightforward, and the observers will often be selected to range actors will just add clutter.

I'm against but can be convinced otherwise.

MatthewMarinets:
Added range indicator when using it manually / hovering over the button:
image

Noticed the hotkey is missing, will fix in a second.

MatthewMarinets:
image

PR #270, opened 2024-08-22
sc2: Adding the dig protoss bases as locations

MatthewMarinets:
Adding Protoss bases as locations in The Dig. Each main base (players 2, 4, and 5) are their own locations.

Pairs with [client PR #292](https://github.com/Ziktofel/Archipelago/pull/292)

## Locations sent when expected when I used cheats to clear it fast. Note the two outposts guarding the western and eastern relics are part of the Northwestern base and Eastern base respectively. Air-wave Pylons are not part of any base.
image
image

## Objectives display after reaching the base:
image

## Reloading the mission after collecting the locations doesn't put the objective back:
image

## Objectives display after reaching the base with mastery locations turned off:
image


MatthewMarinets:
Adding images of the triggers as that would have been really useful on the last PR in this thread for recreating stuff:
image
image
image
(other two protoss base destroyed triggers are almost the same, just with player parameters swapped)
image

PR #173, opened 2024-07-08
Bomber Version 1

MadiMadsen:
Refitting the purifier carrier into the anti ground unit: The Bomber. Primarily Antiground carrier Does high damage but does not discriminate against friend or foe with it's mini bombers!

PR #276, opened 2024-09-05
Purifier carrier rework

MatthewMarinets:
Essentially redid / adapted the changes from PR #173.

* Renamed the unit to the Trireme
* Replaced the purifier interceptors with Bombers, which are larger and drop bombs which do big splash damage but only hit ground

Pairs with [client PR #299](https://github.com/Ziktofel/Archipelago/pull/299)

In some quick testing, I found the damage output vs ground was absolutely ridiculous, but it wiped out friendly units very readily, which I think is a fun balance. If we remove friendly-fire, I'd want to also reduce the damage output.

image
(Corrected this saying they could hit air after taking the picture):
image
image


MatthewMarinets:
Took notes on what was added/removed:
[trireme.md](https://github.com/user-attachments/files/16885987/trireme.md)

Ziktofel:
Damage would be subject to playtesting

Ziktofel:
Snímek obrazovky pořízený 2024-09-05 23-21-50
If I attack an air target, the bomber interceptors move out, and stop near Trireme

Ziktofel:
The thing is that before the unit had just real weapon that handled both interceptors and beam with both Ground and Air targeting, while the beam started to work only if the attack target is air. IIRC it's done this way (and not like multi-lock) due to effect being channeled, thus making the unit busy

Ziktofel:
Use `Assets\Textures\btn-unit-collection-purifier-bomber.dds` asset to display bombers (dummy Trireme weapon, train icon)

MatthewMarinets:
Fixed the icons, good catch.
image

Interceptors also no longer come out vs air.

MatthewMarinets:
Missed the ability in that last one. Now fixed
image

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/EffectData.xml
```diff
@@ -17644,6 +17632,84 @@
...
         <SearchFlags index="OffsetByUnitRadius" value="1"/>
     </CEffectEnumArea>
+    <CEffectCreatePersistent id="AP_ReleaseBombersPatrolCP">
+        <WhichLocation Value="CasterUnit"/>
+        <EditorCategories value="Race:Protoss"/>
+        <Flags index="RandomEffect" value="1"/>
+        <Flags index="RandomOffset" value="1"/>
+        <PeriodCount value="1"/>
+        <PeriodicEffectArray value="ReleaseBombersPatrolLeftTurnCP"/>
+        <PeriodicEffectArray value="ReleaseBombersPatrolRightTurnCP"/>
```
Ziktofel, line 17642: These effects don't exist

MatthewMarinets: Fixed, copied the effects over from starcoop gamedata. Didn't look too closely to see if interceptor movement effects were reusable, but that didn't seem to be the case at a glance.

MatthewMarinets:
Did a few tests to make sure things weren't broken, and found that they were crazy popular. Three triremes could take down a base, would 4-volley a nexus, etc. I tuned the numbers down a bit, so a volley of bombs is 3 bombs instead of 4, and reduced the base damage from 12 -> 9. Tried again, they're still just as flashy, and probably still too strong.

PR #288, opened 2024-09-21
better ingame UI for archipelago messages

Cristall:
i didnt like the buttons to scroll up/down the messages and increase/decrease the size of the message list ("chatbox"), so i rewrote it in SC2Layout.
grafik

The new features are:
- entire frame can be moved on screen via dragging the header label ("Archipelago Messages" text)
- the frame can be resized in the top/down direction via dragging the small arrow image at the bottom
- the list of messages has a scrollbar and you can use mousewheel to scroll through it (scrollbar only appears if messages dont fit into the frame)
- scrolls down whenever a new message is sent

the arrow texture was the first i found, so feel free to change it.
the size x/y in the "AP_Core_MapConfig_setChatboxPosition" and "init chatbox" functions are useless and could be removed, but i didnt want to go through all the missions to fix the api change (if the changes potentially do not get approved).
i also removed the close/open button, since players can now reduce the size of the new chatbox to only the header and total labels, which should be small enough.

on piercing the shroud i had to move the initial position of the chatbox to the left, as the green screenbuttons would be behind the chatbox, preventing them to be used. it wasnt a big deal since players can just move the chatbox around while in the cutscenes. its new position should be below the objectives even if all are visible.


Ziktofel:
Where's the button to collapse the box?

Cristall:
I did not recreate it as I didn't feel the need to completely hide the message box when you can quickly and easily make it very small or move it out of the way.
If that feature is requested, I can add it back in

Ziktofel:
it's sometimes practical to hide in some missions with a lot of UI clutter like All-in

Cristall:
added it back in, including a tooltip
grafik

PR #293, opened 2024-10-01
Adding Warp Relocate

EnvyDragon:
Adds an ability to all Protoss structures allowing them to teleport to a different location within pylon power after a brief delay.

Level 1: All structures (except cannon/shield battery/khaydarin monolith) require 1 charge to warp; charges are shared globally and you can store up to 3

Level 2: Charge restrictions are lifted

Generator PR: https://github.com/Ziktofel/Archipelago/pull/320


Ziktofel:
Snímek obrazovky pořízený 2024-10-01 18-14-29
Moving automated Assimilator is still mining. Is that intended?

Ziktofel:
Snímek obrazovky pořízený 2024-10-01 18-20-10
Dark Shrine can't move

Ziktofel:
Snímek obrazovky pořízený 2024-10-01 18-24-24
Probes can harvest from an Assimilator that's moving away

EnvyDragon:
Pushed fixes:
- Moved the Dark Shrine rally button up one row so it doesn't block the warp button anymore (then I tested every protoss structure and moved them all to ensure there were no more oversights)
- Added behavior to caster which disables resource abilities and drop-off functionality
- Added behavior to destination unit which kills them (canceling the ability) if power is lost
- Fixed the issue where warping assimilators/orbital assimilators would heal them to full (vitals at destination should now match source unit)
- Fixed the issue where pylon power overlay wouldn't appear if you were trying to warp a unit that doesn't need power by default (you need power for a valid warp location so it makes sense to have it visible)

Ziktofel:
in:
```xml
    <CBehaviorBuff id="AP_AutoHarvestVespeneProtoss">
        <EditorCategories value="AbilityorEffectType:Units"/>
        <Period value="2.1"/>
        <!--Period value="1.68"/-->
        <PeriodicEffect value="AP_AutoHarvestVespene4"/>
        <RemoveValidatorArray value="AP_AutoHarvesterFinish"/>
        <InfoFlags index="Hidden" value="1"/>
        <BuffFlags index="DisableWhileUnderConstruction" value="1"/>
        <DisableValidatorArray value="AP_AutoHarvesterHarvestProtoss"/>
        <DisableValidatorArray value="AP_HaveCinematicModeDisabled"/>
    </CBehaviorBuff>
    <CBehaviorBuff id="AP_AutoHarvestVespeneProtossRich">
        <EditorCategories value="AbilityorEffectType:Units"/>
        <Period value="2.1"/>
        <!--Period value="1.68"/-->
        <PeriodicEffect value="AP_AutoHarvestVespene8"/>
        <RemoveValidatorArray value="AP_AutoHarvesterFinish"/>
        <InfoFlags index="Hidden" value="1"/>
        <BuffFlags index="DisableWhileUnderConstruction" value="1"/>
        <DisableValidatorArray value="AP_AutoHarvesterHarvestProtoss"/>
        <DisableValidatorArray value="AP_HaveCinematicModeDisabled"/>
    </CBehaviorBuff>
```
add a validator if you want to prevent mining while relocating

EnvyDragon:
> add a validator if you want to prevent mining while relocating

Looks like what was happening is that the existing fix (the relocate behavior disabling "resource" class behaviors) was inconsistent, because in testing I was still able to catch examples of relocation disabling the harvesting effect, but I'm not really sure why. 

I think it had something to do with the fact that I was generally testing the nexus's relocation at the same time, which would result in briefly not having a finished nexus (a requirement for autogas, as I just learned), so maybe there was some order-of-operations thing going on.

Either way, I just added the validators. I figured a behavior was best on its own for expandability but it's not a huge issue. Consistency is better.

PR #296, opened 2024-10-09
Various small QoL fixes for Zerg

EnvyDragon:
Roachling (and other Broodling-ish units):
- Reduced unit priority to 5 (from 14)

Merc Zergling:
- Swarmling upgrade now increases spawn quantity from 4 to 6

Ravager:
- Corrosive Bile now continues cooling down while burrowed

Swarm Queen:
- Autocast heal no longer targets timed-life units (ex. roachling, infested terran)
- Should now correctly perform cast animation when healing


Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/UnitData.xml
```diff
@@ -31,6 +31,10 @@
         <SelectAlias value="Broodling"/>
         <AIEvaluateAlias value="Broodling"/>
     </CUnit>
+    <CUnit id="BroodlingDefault">
+        <!-- Override  -->
+        <SubgroupPriority value="5"/>
+    </CUnit>
```
Ziktofel, line 37: Instead of overriding this unit you could re-parent affected units to `AP_BroodlingDefault`

PR #308, opened 2024-11-10
sc2: Adding Smash and Grab Kerrigan base location

MatthewMarinets:
Pairs with [client PR #338](https://github.com/Ziktofel/Archipelago/pull/338)

Adding a mastery location for destroying Kerrigan's base in Smash and Grab. Testing pending.

## GUI trigger changes
image
image
image
image

## In-game changes
image
image
image

Note Kerrigan still gets occasional spawned in zergling/hydra waves and creep tumors. This doesn't seem to be enough to get through most defenses, so mostly serves to protect your new base from protoss attack waves. Note it means there's still a non-zero risk of losing after this. If the main protoss base is broken, a large zerg wave will spawn.

I got the check legit with raven+weapon, hellion, and turret, with +1 air attack ups. The high-ground outside the zerg base is incredibly strong to camp on, and basically any powerful air dealer + turret + a wall can probably do it.

## Known issue
Currently, this _mostly_ works, except that you have to kill one additional protoss building after the zerg is eliminated to properly get the check. I'll likely update the trigger condition to allow for, say, 3 zerg buildings remaining to reduce the need to hunt extra buildings and to get around this issue.


MatthewMarinets:
Fixed the trigger. It seems that when the trigger runs conditions, the triggering unit from a "unit dies" event isn't dead yet as far as unit counts are concerned. So now the trigger checks that:
* 3 or fewer zerg buildings remain
* No "prevent reveal" buildings remain (hatch/lair/hive)
* OR 1 "prevent reveal" building remains and the unit killed is a "prevent reveal" building

image

Tested this on terran/protoss/zerg with cheats this time and things behaved as expected, both when the last building killed was a hatchery and when it wasn't.

PR #313, opened 2024-11-19
Add biomechnical stockpiling upgrade

itsjustbones:
reduces max charges for infested mech buildings to 2, and adds an upgrade to increase it to 4 (old base was 5)


itsjustbones:
I changed it from 4 to 5

PR #318, opened 2024-11-25
Trireme bomber adjustment

EnvyDragon:
Goal of changes: Make the Trireme bombers feel a little more like they're doing actual bombing runs; adjust their damage profile slightly vs smaller units (slightly weaker vs smaller stationary units, but able to deal a non-zero amount of damage vs small moving units). Also make the visual/audio effects a little more gentle to reflect the reduced impact of individual bombs.

Superficial changes (because the effects were previously scaled for the co-op version iirc, so they were built around a much bigger impact):
- Radius of explosion VFX reduced to 60% of normal
- Volume of launch and explosion SFX reduced slightly

Changes to bomber drone:
- Added minimum speed of 4 (so they'd quit stopping in midair, which means they randomly concentrate their bombs on one location)
- Increased LoiterInnerRadius 2->3 so they move a little further before turning
- Increased weapon range 0.1->0.75 so they start dropping bombs slightly ahead of their target
- A few tweaks to rangeslop and arc so they don't constantly try to retarget

Changes to bombs:
- Period between drops increased 0.125->0.5 (spreads out damage a bit)
- Damage radius profile changed: 
-- Full damage at 0.5 radius -> Full damage at 0.25 radius, half damage at 0.5 radius, quarter damage at 1 radius (so you do a small amount of damage more frequently, but direct hits still hurt just as much)
- Mover changed from "Guidance" to "Throw" with increased speed (so the bomb just drops, and it drops much faster)

Observations in testing:
- Still basically the same vs structures, though _slightly_ weaker vs 2x2 structures, and slightly stronger vs tightly grouped structures
- Much weaker vs, say, individual zerglings minding their own business; now much more effective at destroying grouped units however, and now less ineffective vs fast-moving units (they deal chip damage rather than nothing)

### DIFF
--- a/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/EffectData.xml
+++ b/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/EffectData.xml
@@ -17968,9 +17968,8 @@
         <PeriodicValidator value="CasterNotDead"/>
         <PeriodicEffectArray value="AP_CarrierPurifierBomberLM"/>
         <PeriodicPeriodArray value="0"/>
-        <PeriodicPeriodArray value="0.125"/>
-        <PeriodicPeriodArray value="0.125"/>
-        <PeriodicPeriodArray value="0.125"/>
+        <PeriodicPeriodArray value="0.5"/>
+        <PeriodicPeriodArray value="0.5"/>
         <PeriodicOffsetArray value="0,-0.1,-3"/>
     </CEffectCreatePersistent>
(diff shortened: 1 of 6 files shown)


MatthewMarinets (approved):
Changes and intent behind changes in the description look good to me.

I notice the friendly-fire damage was unchanged. The intent of the old design was that it was same splash radius as anti-enemy, but 1/3 damage and _not_ increasing with upgrades. May be better to change the area to 0.25 to match innermost anti-enemy splash radius, or 3 dmg at 0.25 range and 1 damage at 0.5 range or something?

PR #320, opened 2024-11-25
Mm/broodling upgrade

MatthewMarinets:
Adds a zerg global upgrade that increases the amount of broodlings that are released from buildings on death.

Pairs with [client PR #352](https://github.com/Ziktofel/Archipelago/pull/352)

# Before upgrade
image

# After upgrade
Crawlers have an icon:
image
Rooted crawler:
image
Evo chamber:
image
Uprooted crawler:
image
Hatchery:
image


MatthewMarinets:
image
Also updated the yggdrasil description as I didn't update strings when giving them baseline transport in PR #316

PR #321, opened 2024-11-26
sc2: Added Pygalisks

MatthewMarinets:
Adding pygalisks.

pairs with [client PR #357](https://github.com/Ziktofel/Archipelago/pull/357)

Currently adding them on the Predator Nest, but I'm not completely happy with it, because the whole point is they're supposed to be accessible. Gave them a 30s cooldown and uncapped count by default, though that means merc upgrades can't realistically affect them. Currently 6 for ~~300 minerals (50 minerals per)~~ 450 minerals (75 per).

Edit: Trying them out on a few maps, I found I was very constrained by cooldown times with no real recourse. To address this, I moved the CD restriction from being player-bound to unit-bound (unit being Predator Nest), so multiple Predator Nests can be used to get better production. Still not happy with this as it's the only unit that behaves this way.

I buffed their stats slightly from Piercing the Shroud:
* Damage +4 (4->8)
* Life +25 (50->75)
* Armor +1 (2->3)

Also gave them burrow and a chibified version of the Ultra's old Ram attack, at about half the attack speed of their vs units attack but 25 damage per attack.

I like the identity of a small, cheap, but durable army that has low DPS. They still melt vs anything that does more than zergling damage per attack lol.

## TODO
- [x] Proper Unlock Trigger
- [x] Some upgrades
  - [x] Ducal Blades (splash damage)
  - [x] Stim (they are a dominion experiment)
  - [x] Combat Carapace (+25 health, mashup of Combat Shield+Organic Carapace)


MatthewMarinets:
image
image
image

Changed around the wireframes and buttons a bit so you can more easily tell them apart from ultras in the command-card. Not fully happy with the button, it's kind of the wrong colour, but all the other ultralisk buttons are even more off (and the baseline one looks huge)

MatthewMarinets:
Tested with `/send` commands, everything unlocked properly (even the Predator nest on getting Pygalisks).
image

I'm not fully happy with them on the predator nest; it would be nice to see a different way to build them, but it will do for now until we agree on a better solution.

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/UnitData.xml
```diff
@@ -23435,6 +23435,126 @@
...
+        <LifeMax value="75"/>
+        <LifeArmor value="3"/>
+        <LifeRegenRate value="0.2734"/>
+        <LifeArmorName value="Unit/LifeArmorName/ZergGroundArmor"/>
+        <ShieldArmorName value="Unit/ShieldArmorName/ZergWildMutation"/>
+        <Mover value="Burrowed"/>
+        <Sight value="6"/>
+        <Food value="-1"/>
+        <CostCategory value="Army"/>
+        <CostResource index="Minerals" value="50"/>
```
Ziktofel, line 23530: Snímek obrazovky pořízený 2024-11-28 17-46-10
Should match the unburrowed form

Ziktofel (changes requested):

Ziktofel (changes requested):

Ziktofel:
Snímek obrazovky pořízený 2024-11-28 21-44-19
No HP upgrade, no upgrade icons

PR #332, opened 2024-12-04
sc2: Updating several hotkey and glossary categories for infested and royal guard units

MatthewMarinets:
Got all the Royal Guard units showing up in the correct category, and updated all zerg infested structures and units to the zerg category. The zerg category is so full at this point that it desperately needs a split, but I'm leaving that for another day.

image
image
image

The two no-icon buildings in the zerg tab are infested missile turret and bunker. I'm not sure why they don't get icons; I looked around and didn't find anything out of place in actor data, and the rootable anywhere variants probably aren't an issue or I'd expect spine and spore to display the same problem.


MatthewMarinets:
Checked the glossary, and there's a ton of duplicate entries in the terran one in particular. I think it's related to different states for a unit, though there were cases of e.g. emperor's shadow having the issue. All the ground categories are completely full regardless, so I won't worry about it too much. That also implicitly fixes the Queen glossary issue ([#147](https://github.com/Ziktofel/Archipelago/issues/147) ) as it's no longer observable.

PR #335, opened 2024-12-08
Mm/royal guard upgrades

MatthewMarinets:
Progress so far following RG feedback and discussion in the discord:

* Mengsk RG don't benefit from +attack/armour upgrades, so the current upgrades are essentially double-dipping on power, for units that are already strong enough without
* Some upgrades _multiply_ the effects, which is just unnecessary
* Goals:
  * Remove multiplicative effects
  * Remove armour from leveling
  * Half damage bonus from attack ups

Another possible route would have Royal Guards don't benefit form w/a upgrades at all, similar to Mengsk.

### DIFF
--- a/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
+++ b/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
@@ -8049,8 +8049,6 @@
 <!--                <DamageDealtFraction index="Melee" value="0.25"/>-->
 <!--                <DamageDealtFraction index="Ranged" value="0.25"/>-->
 <!--                <DamageDealtFraction index="Splash" value="0.25"/>-->
-                <LifeArmorBonus value="1"/>
-                <ShieldArmorBonus value="1"/>
                 <VitalMaxArray index="Life" value="200"/>
             </Modification>
         </VeterancyLevelArray>
@@ -8059,8 +8057,6 @@
 <!--                <DamageDealtFraction index="Melee" value="0.25"/>-->
 <!--                <DamageDealtFraction index="Ranged" value="0.25"/>-->
 <!--                <DamageDealtFraction index="Splash" value="0.25"/>-->
-                <LifeArmorBonus value="1"/>
-                <ShieldArmorBonus value="1"/>
                 <VitalMaxArray index="Life" value="200"/>
             </Modification>
         </VeterancyLevelArray>
@@ -8352,7 +8348,6 @@
 <!--                <DamageDealtFraction index="Melee" value="0.25"/>-->
 <!--                <DamageDealtFraction index="Ranged" value="0.25"/>-->
 <!--                <DamageDealtFraction index="Splash" value="0.25"/>-->
-                <LifeArmorBonus value="1"/>
                 <VitalMaxArray index="Life" value="70"/>
             </Modification>
         </VeterancyLevelArray>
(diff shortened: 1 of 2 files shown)


MatthewMarinets:
I'm not going to manage a complete test of this alone. Covering the case of level 1 attack/armour on the bulwark company:

Context: 0/0 goliath is 125 HP, 1 armour, 8/16, 18

| Level | Upgrade level | health | armour | damage 1 | damage 2 |
| --- | --- | --- | --- | --- | --- |
| 0 | 0/0 | 250 | 1 | 12/24 | 27 |
| 1 | 0/0 | 320 | 1 | 30 | 33.75 |
| 2 | 0/0 | 390 | 1 | 36 | 40.6 |
| 0 | 1/1 | 250 | 2 | 13/25.5 | 28.5 |
| 1 | 1/1 | 320 | 2 | 31 | 35.25 |
| 2 | 1/1 | 390 | 2 | 37 | 42 |
| 0 | 2/2 | 250 | 3 | 14/27 | 30 |
| 1 | 2/2 | 320 | 3 | 32 | 36.75 |
| 2 | 2/2 | 390 | 3 | 38 | 43.5 |
| 0 | 3/3 | 250 | 4 | 15/28.5 | 31.5 |
| 1 | 3/3 | 320 | 4 | 33 | 38.25 |
| 2 | 3/3 | 390 | 4 | 39 | 45 |

image

Ziktofel:
Need to play actual missions in a normal game myself to make up my mind first. Maybe I'll figure out something better as giving only 0.5 dmg/level isn't the thing that's expected in SC2 (W/A upgrades add about 10% of the base dmg, but the minimum dmg added for direct hits is 1 dmg per level)

BTW the multiplicatives are Vanadium bonuses to scale properly with levelups (A levelup adds 25% bonus HP)

MatthewMarinets:
Personally, I'm leaning more and more in favour of just not having RG benefit from w/a at all. Bones tried to say they should be affected, which is why I settled for halfway.

The feedback on them so far has been that they feel good and shockingly balanced for a co-op unit at baseline, and the exp steal / levelup mechanic provides a neat counterforce to the insane powercreep of getting too many items. However, stacking additional items and power on top of that then makes them overtuned. This is most apparent in the case of the bulwark company, which benefits from all the really strong goliath upgrades meant to take a mediocre unit to a good unit, that now take a good unit to an insanely overpowered unit. Attack/Armour upgrades are just one component to them scaling too hard.

I think ultimately best route forward is to remove all or most effects from baseline units from affecting RG, and to remove w/a effects on RG entirely. Cutting damage upgrade by half and removing armour from level up already feels like a half-measure and I regret compromising on that before even getting to testing.

Ziktofel:
The thing is that if they won't benefit from regular unit upgrades is that they'll lag behind standard units (and you'd make 2 regular dudes instead)

Also there's a ton of shared code.

Also the benefit from W/A upgrades is mainly due to L5 W/A upgrades existing

Still need to a play game myself. If Bulwarks are the problematic ones, I can look into them

MatthewMarinets:
I think it's a totally fine dynamic to sometimes make regular units over RG units -- it's healthy for other items / upgrade items / mission duration to change what units work better. I'd call it unhealthy if it was strictly better to always build RG if they were available.

MatthewMarinets:
After getting feedback from more players and further discussions in dev chats, we're probably going to go a different direction:
* RG leveling will still give armour
* RG will not benefit from weapon/armour upgrades
* RG will not benefit from base unit upgrades. For example, Bulwark Companies will not get health from Goliath Shaped Hull
  * Unique effects or interesting effects like Jump Jets, transformation speed, or extra abilities will likely get separate items for RG
* Some specific Royal Guard may need a balance pass, or to be placed in OP items. The ones currently on my radar are Aegis Guard and Bulwark Companies
  * Balance pass may involve changing stats, changing levelling bonuses (possibly giving abilities that currently come from base unit items), or making them more expensive
  * Another possibility is to make getting the production buildings harder, say by making them more expensive or making them take much longer to build
  * As this won't change the item description and is unlikely to affect logic, this change may come after feature freeze or after the client merges to main

With all this in mind, I'm closing this PR and will make a separate PR to remove benefit from weapon/armour upgrades and make related client/description changes. The client PR is already in draft as [client #409](https://github.com/Ziktofel/Archipelago/pull/409), it will link to the data PR when it's ready.

PR #338, opened 2024-12-12
sc2: Adding bile launchers

MatthewMarinets:
Added bile launchers.
Pairs with [client PR #372](https://github.com/Ziktofel/Archipelago/pull/372)

Things are ported almost exactly the same as co-op Zagara, with a minor nerf to artillery ducts: it gives +8 range instead of co-op +10.

Haven't playtested for balance too much, just a little on S&G and Safe Haven. They were very much able to snipe attack waves before they started coming in, and the nexuses on SH. However, as soon as an attack wave rolled in, they were very extremely vulnerable and required manual micro to be effective. Pulling of a skillshot felt very good, but was very easy to screw up. Defensively, they definitely would like a wall / screening, and offensively they're very powerful but are hampered by their reliance on creep, expensive gas cost, and slow build time (40s).

image
image
image


MatthewMarinets:
Added some additional passive icons. On the bile launcher itself added one for broodling packing and  (left off creep stomach as non-crawlers also leave it off).
image
image

image
image

PR #339, opened 2024-12-14
sc2: Adding Mothership WC and Supplicant Zenith Pitch upgrade

MatthewMarinets:
Adding supplicant upgrade:  Zenith Pitch. Allows Supplicants to target air.
Pairs with [client PR #375](https://github.com/Ziktofel/Archipelago/pull/375)
Progress on #152 

Note I gave the mothership a behaviour to get selecting it to display power fields. However, I couldn't find a way to get the upgrade to _remove_ this behaviour, so motherships will just always show pylon power fields.

## Mothership Pictures
mothership_slow
image
image
image

## Supplicant pictures
supplicant_zenith_pitch
image
image
image


Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
```diff
@@ -4688,6 +4688,24 @@
...
+        <EditorCategories value="AbilityorEffectType:Units"/>
+        <DisableValidatorArray value="AP_NotHaveMothershipTaldarimIntegratedPower"/>
+        <DisableValidatorArray value="AP_CasterLocationNotHasPower"/>
+        <Modification MoveSpeedMultiplier="0.2"/>
+    </CBehaviorBuff>
+    <CBehaviorPowerSource id="AP_DisplayPowerFieldMothership">
+        <InfoFlags index="Hidden" value="1"/>
+        <InfoIcon value="Assets\Textures\btn-building-protoss-pylon.dds"/>
+        <EditorCategories value="Race:Protoss,AbilityorEffectType:Structures"/>
+        <DisableValidatorArray value="AP_NotHaveMothershipTaldarimIntegratedPower"/>
```
Ziktofel, line 4704: Editor Warning:
`Warning: XML: Z:\hdd\home\venca\software\battle-net\drive_c\Program Files (x86)\StarCraft II\Mods\ArchipelagoPlayer.SC2Mod\Base.SC2Data\GameData\BehaviorData.xml(4704,9) : Unable to find field DisableValidatorArray  Source: Mods/ArchipelagoPlayer.SC2Mod  Entry: AP_DisplayPowerFieldMothership  Scope: CBehaviorPowerSource`

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
```diff
@@ -4688,6 +4688,24 @@
             <Chance value="1"/>
         </DamageResponse>-->
     </CBehaviorBuff>
+    <CBehaviorBuff id="AP_MothershipTaldarimMovementNerfOutsidePower">
+        <Alignment value="Negative"/>
```
Ziktofel, line 4692: Medic's Restoration removes that. Recolor the info icon instead. I've removed the alignment on behaviors those removal by Restoration cause bugs

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
Ziktofel: I believe this is the root of your bug as the behavior isn't a `CBehaviorBuff`. You can, however remove the behavior with a trigger (There's a trigger that removes Ursadon's `CBehaviorLeash` on being derpfestor'd)
You want to remove the behavior on following events: An upgrade is being given or a unit is trained

Ziktofel (changes requested):

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
Ziktofel: Alternatively, you could have a second behavior that's acting as a controller and when exiting, do the remove behavior effect that removes the power source

Ziktofel: BTW found that Power source behavior can use requirements:
```xml
    <CBehaviorPowerSource id="AP_WarpPrismPowerSource">
        <InfoFlags index="Hidden" value="1"/>
        <EditorCategories value="Race:Protoss,AbilityorEffectType:Structures"/>
        <Radius value="3.75"/>
        <PowerLink value="PowerSource"/>
        <Requirements value="AP_NotHaveWarpPrismWarConfiguration"/>
    </CBehaviorPowerSource>
 ```

File: Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml (hunk 2 above)
Comment (MatthewMarinets) [reply Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml:4692]: The medic can remove buffs with the Permanent flag? Feels like a bug on the Medic's end. Will test it out, though.

Comment (MatthewMarinets) [reply Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml]: Removed the negative flag, tested and medics could no longer restore it away. Also tried tinkering with more filters on restoration, but that didn't work so undid those changes and just went with avoiding the negative flag.

File: Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml (hunk 1 above)
Comment (MatthewMarinets) [reply Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml]: So I've tried
* Requirements
* DisableValidator
* RemoveValidator
* Removing straight from the upgrade's EffectArray
* Having another behaviour remove it with BehaviorLinkDisableArray

Some of those removed the behaviour icon, but none removed the power field visuals. As far as I can tell, it's the presence of a link to this behaviour type at all that works.

Checking the triggers, `UnitBehaviorRemove` seems to target a particular unit rather than a unit type. I'll try tinkering with a behaviour chain before I go the trigger route, as triggers likely means a persistent trigger that enables / disables behaviours continuously.

Comment (MatthewMarinets) [reply Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml]: I think I figured out how to approach this. Buried all the way in core is this:
```xml
    <CActorPower default="1">
        <PowerSource value="##id##Source"/>
    </CActorPower>
```
Which seems to be the root of the display woes. I should be able to trace this through actordata and deal with the visuals from there.

MatthewMarinets:
The power field display should be working as intended now. Without the upgrade, Pylon power fields display when the mothership is selected. When the upgrade is acquired, this behaviour goes away.

works_mothership

This was accomplished by a unit morph to an alternate mothership unit. As far as I can tell, PowerSource layer display is controlled by the game binary and is not modifiable from actordata. It displays if a unit is selected and that unit's behaviorArray includes either a `CBehaviorPowerSource` or `CBehaviorPowerUser` with `DisplayPower` flag set to 1. So the only way to control it in data is with the classic unit morph trick we've used for cargo capacity and rootable footprints.

Did a check afterwards. The actordata seemed to work fine, stargate build display still works the same, and I fixed the mothership requirement to also account for the new mothership variant.

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
MatthewMarinets: Opened things in the editor to check, all error messages related to this PR should be cleaned up
image

Ziktofel:
Snímek obrazovky pořízený 2024-12-15 11-50-47
Snímek obrazovky pořízený 2024-12-15 11-53-00
Armor upgrades don't work, only update the icon, not unit stats if the Mothership is upgraded

PR #343, opened 2024-12-23
New unit: Skylancer, Skylord rework

Ziktofel:
# What's changed
- Changes Skylord to use 4 interceptors, doing 2*18 air damage and 2*5 ground damage (ground attack is the same as regular interceptor's)
- The old Skylord laser attack is moved to a new unit - Skylancer, that's now a Tempest variant, otherwise behaves like Skylord before
- Skylancer has a new upgrade that doubles its attack speed but overheats while attacking (stops at original attack speed)

# Motivation
- Bring some consistency into the game as the upgrade paths for Carrier class are somehow inconsistent (Hull of Past Glories) but Graviton Catapult is impossible to fit the old Skylord
- Unit shapes make their behavior more predicable
- Skylord more matches its name (big air unit that's a great anti air)

# Misc
See generator PR https://github.com/Ziktofel/Archipelago/pull/380


MatthewMarinets:
As discussed in the discord, I don't think we _want_ another carrier unit, and I don't think it's important that Graviton Catapult applies to the whole class. Most people are against this change. I am neutral on renaming the Skylord to Skylancer; I like the old name largely because I'm used to it.

I'm against the model change for the Skylord as I designed the unit with the model in mind. The inspiration came from the sc1 opening art of carriers frying the surface of Chau Sara, so it feels weird to put the laser on a Tempest model. The precise laser model seemed to wrap around the contours of the Skylord model. And the larger form of the carrier-type hull gives the player impression of bulk, which matches with the unit's large health pool, which in turn felt appropriate to encourage the player to move the unit forward to get the most of the linear splash.

The increased attack rate to start is an interesting idea, but I'm not that big a fan personally. The slow attack rate of the Skylord would often lead to situations where the player would intentionally stagger the attack cycles of multiple skylords to make sure there was one in a small group that could react to a new anti-air attacker approaching. Having a variable attack rate makes this more difficult. A short-range blink, similar to a mothership, was requested and something I'd like to see on the skylord; the blink could reset the attack cycle if we want some way to increase the attack rate. Note the damage numbers and attack rate have already gone through a balancing / tuning phase following feedback, I'd expect another such phase would probably have to follow if we applied a general attack speed buff to the Skylord. Someone also pointed out that giving more power at the start of an engagement is a little odd on a unit that can just teleport out at any time.

If it's a high priority to add a new tempest-class unit or add more Protoss items, I can look into making a new Tempest variant or reworking the Warp Ray or splitting up / adding protoss upgrades to get more items on request; there are some tracker changes I was working on otherwise. I don't think it's that urgent, though, and I don't think it's worth adding repeat units or more generalist units like this when there isn't urgency.

There were many other comments in the discord, but suffice to say I don't think anyone really wants these changes in general. To me the largest point against is that I don't see any kind of artistic or design intent here -- it feels like moving things around for the sake of having another carrier, when we already have two carriers and that's already more than we need for a randomizer. The justification of having graviton catapult apply to more units feels unwarranted, as we generally want more items that apply to fewer units, with sentiment suggesting we should break protoss items up to apply to individual units rather than have the same items apply to move units. Additionally, a carrier variant, even one specialized into anti-air, that can still hit ground is enough of a generalist that it detracts from the randomizer experience. The most interesting situations are when the player is saddled with only niche/specialist units and must cobble them together into a strategy; this situation lapses as soon as the player collects 1~2 generalists. Protoss is already overloaded with generalists that can handle most situations, it does not need another one that is survivable, low-micro, and can hit both ground and air.

PR #355, opened 2025-01-15
sc2: Adding sensor tower assistive targeting upgrade

MatthewMarinets:
2 new upgrades for the sensor tower to give them some use.
Pairs with [client PR #394](https://github.com/Ziktofel/Archipelago/pull/394)

## Assistive targeting
+1 range for attacking buildings. Does not affect spell towers like HME or Psi Disrupter. Tested that it increases the range of bunker shrike turrets, but it did not increase the range of units in the bunker.

While testing it seemed that projectile accelerator didn't seem to give extra range, at least for manually targeting with the attack button. I could issue an attack order no further with the upgrade than without it.

image
image
image
image
image

## Enhanced Vision
+10 sight range, +5 radar range, improves assistive targeting. Kind of trashy, but I could see it synergizing with super-range tanks or calldowns/abilities. At least it's on-theme
image

Before:
image

After:
image


MatthewMarinets:
When testing the bunker, I found that projectile accelerator had no effect. Was a simple case of a missing validator, quick fix. Tested with this setup:
image

Without the fix, the marine in the bunker could hit SCV 2 but not 3, and getting projectile accelerator didn't affect that.

With the fix, Projectile accelerator let the marine hit SCV 3.

PR #362, opened 2025-01-20
sc2: Updating colonist stats on raceswapped infested so they're not useless

MatthewMarinets:
This is responding to feedback from multiple people. The colonists on raceswapped infested just felt bad:
* They had no DPS
* They outranged bunkers, meaning they didn't tank for you
* They generally didn't do anything

Updated their stats to be a little closer to infested colonists without duplicating it entirely:
infested civilian:
35 health
8 damage / 1.2s = 6.15 DPS
melee

vanilla civilian:
30 health
5 damage / 7 + (2~5)s = 0.417 ~ 0.55 DPS
7 range

Updated civilian:
35 health
11 damage / 2 + (0~1) = 3.67 ~ 5.5 DPS
3 range

image

### Testing
Played through the start as Terran and Protoss. The civilians felt _much_ better -- they didn't carry the mission, and required support to make any progress vs tanks or bunkers, but they tanked and helped take down those obstacles with support / sniping. The health buff affected protoss because they start with ascendants with sacrifice, so it meant more energy.

### Design thoughts
I aimed for a DPS that was slightly lower than infested civilians for a few reasons:
* A bit of range probably meant a higher likelihood to connect and do chip damage before dying
* More range means more concaves / simultaneous attacks when not dying
* Higher attack at slower rate means DPS is less impacted by armour

That said, it was still a little on the low side. I think damage can be further bumped 11 -> 12 if people still find it too low.

### Other observations
Race-swapped infested is probably the buggiest / least polished mission I've seen so far. I've listed bugs in the feature freeze goal list. Suffice to say, there's lots of display bugs in tooltips and objective text which I have _not_ fixed in this PR.

### DIFF
--- a/Maps/ArchipelagoCampaign/HotS/ap_infested.SC2Map/Base.SC2Data/GameData/EffectData.xml
+++ b/Maps/ArchipelagoCampaign/HotS/ap_infested.SC2Map/Base.SC2Data/GameData/EffectData.xml
@@ -1,4 +1,8 @@
 <?xml version="1.0" encoding="us-ascii"?>
 <Catalog>
     <CEffectDamage id="MagnisHammerDamage"/>
+    <CEffectDamage id="ThrowMolotovDamage">
+        <!-- Override -->
+        <Amount value="11"/>
+    </CEffectDamage>
 </Catalog>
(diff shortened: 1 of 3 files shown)


Ziktofel:
As the civilians do a bit of splash damage, the lower base dmg is OK

PR #363, opened 2025-01-20
SC2: Add Instigator Reconstruction

Snarkie:
Adds a new item to the Instigator (Stalker Purifier): 
### Reconstruction
grafik
Works similar to the Sentinel variant. Upon taking fatal damage, you are revived after a couple of seconds, and you cannot be revived anymore for a long cooldown. Instigators can reduce the cooldown by using blink.
GIF_20 01 2025_12-53-06

Includes a custom icon. Note that it requires a trigger in the Player mod. Also note that the reversed animation is quite a mess in the actor events. Looks good tho :)

## How was this tested?

Local test map for assets and functionality. Local campaign for item functionality and description. Tested interaction with some other items, notably Guardian Shell.

[Client PR](https://github.com/Ziktofel/Archipelago/pull/405)


Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/Lib15EF4C78.galaxy
```diff
@@ -6686,6 +6686,30 @@ void lib15EF4C78_gt_SupplicantWarpCancel_Init () {
...
+bool lib15EF4C78_gt_InstigatorReconstructionCooldownRecovery_Func (bool testConds, bool runActions) {
+    // Automatic Variable Declarations
+    // Actions
+    if (!runActions) {
+        return true;
+    }
+
+    if ((UnitHasBehavior(EventUnit(), "AP_StalkerPurifierRecentlyReconstructed") == true)) {
+        UnitBehaviorSetDurationRemaining(EventUnit(), "AP_StalkerPurifierRecentlyReconstructed", (UnitBehaviorDuration(EventUnit(), "AP_StalkerPurifierRecentlyReconstructed") - 3.0));
+        UnitBehaviorAddCooldown(EventUnit(), "AP_StalkerPurifierRecentlyReconstructed", "Behavior/AP_StalkerPurifierRecentlyReconstructed", -3.0);
```
MatthewMarinets, line 6701: Is this robust to blinking with <3s remaining on the cooldown? I'd expect it to be fine, but just double-checking nothing weird happens if it goes negative.

Snarkie: Yes, I tested this with a value > 240

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/ActorData.xml
Snarkie, line 1: Found a model flag, that allowed me to remove the DestroyParticle messages, which also cleaned up the visuals a bit. Should be better performance this way and removes some of the actor message mess.

Still, faking the reversed animations takes some ugly actor work, I don't know a way to avoid that.

PR #364, opened 2025-01-21
sc2: Balance adjustments

MatthewMarinets:
Crossing off two items on the list from feedback:
* Lessening Blood Hunter WC nerf to just 50% attack speed (see #210 for original values)
* Reducing Reaver Resource Efficiency to something more reasonable.
  * Base cost: 300/200/6, old RE: -100/-100/-2
  * -50/-25/-1 is about the most generally good / average amount I could think of, I could get behind focusing more on one resource, so maybe -25/-25/-2 or even -100/0/-2 to copy the ultralisk one. I don't think it's safe to both reduce gas and food cost by a lot, and I don't think we want any RE going past -50 gas except Predator.

Pairs with [client PR #408](https://github.com/Ziktofel/Archipelago/pull/408)

## Blood Hunter
image
image
image

## Reaver
image
image


MatthewMarinets:
Tossing in Vanguard tooltip fix as Chuck reported it just a few hours after I'd seen the blood hunter issue.
image
image

Ziktofel:
Noted the Reaver into the Google doc

PR #366, opened 2025-01-23
Skylord Jump

MatthewMarinets:
Changing out Skylord Hyperjump (BC tactical jump-style teleport) for Skylord Jump (blink-style teleport). Collab with MindHawk, he volunteered some time and got most things changed himself. I tested on top of that.

There was an editor save involved that needed filtering out. I kept a select few changes for the boy scout principle, to an extent that I felt it didn't add too much clutter to the review. Mostly indentation changes.

Pairs with [client PR #410](https://github.com/Ziktofel/Archipelago/pull/410)

## Testing
image
image

Tested in-game, it had great gamefeel as the skylord appears at a stop and may thus attack instantly.


Ziktofel:
Couldn't be the button named `Blink` instead? As it does exactly the same as other Blinks (and Tal'darim Mothership has Blink too)

MatthewMarinets:
> Couldn't be the button named `Blink` instead? As it does exactly the same as other Blinks (and Tal'darim Mothership has Blink too)

This is more similar to the Hyperion teleport ability in With Friends Like These, which is called "Jump"

PR #369, opened 2025-01-25
Royal Guard weapon/armour rework

MatthewMarinets:
After a lot of discussion, we know that Royal Guard units are fundamentally on the right track, but are too strong in certain conditions. Long-term this will probably involve a few rebalances and removing the effect of most base unit items on RG. In the next dev cycle, RG units may gain some upgrade items of their own.

The goal of this PR is to:
* Remove the effect of weapon/armour upgrades on Royal Guard units; they will only gain those stats from levels
* Rework the internals of older RG weapons to use accumulators / be easier to work with
  * If a base unit upgrade affects the RG weapon such that it's harder to work with, remove the effect of the upgrade (otherwise leave it for the next PR)
* If the royal guard weapons are way stronger than co-op Mengsk, balance them back down to Mengsk levels (or lower)

Work in progress.
Pairs with [client PR #409](https://github.com/Ziktofel/Archipelago/pull/409)

## Battlecruiser / Pride of Augustgrad
* Switch weapon to use accumulators
* ATX was complicating the weapon arrangement, and co-op BCs had it baseline anyway; removed the effect (attack while moving is now baseline)
* Set ATS damage to co-op level (15 -> 12)
* Updated tooltip to clarify how anti-air damage worked (75% AG damage)
image

## General
* Halved xp from infested civilians and infested marines. Outbreak was leveling RG hilariously fast


MatthewMarinets:
I'm told there's a way to avoid `AttributeFactor` with accumulators that target particular bonuses; will experiment when I'm back in town in a couple evenings.

MatthewMarinets:
After messing with accumulators for a few days, it seems like they don't play nice with the weapons display with attribute bonuses. Changing things over the the vanilla Mengsk behaviour of just a flat +25% damage buff per rank. This shows up differently in the display:
image
image
image

I've also updated the medivac transport hook display and tested it's doing the correct amount of damage
image

MatthewMarinets:
Doing the same for battlecruisers. Should be able to get most of the others done later this evening.
image

MatthewMarinets:
RG Marine weapon:
* Splash effect was missing all the normal missile,hidden,dead,invulnerable flags, so theoretically could be used to take out the Supernova fire wave
* Lowered splash radius at level 3 as they were splashing around buildings while testing
* Lowered their base health 100->75. They gain +25/level, and 150 feels like a rounder number while still being entirely fair at 175 minerals 2 food
* There were some complicated weapon switching behaviours going on that seemed unnecessary; I removed all the switching logic and the extra weapon and just added the +2 range to the veterancy behaviour (now displays same as Siege Tank bonus range)

image
image
image
image

MatthewMarinets:
## Marauder
* Just cleanup, no real stat changes
* Removed the weapon swapping behaviours
* The stats on RG are insane, the 300 HP shield definitely needs a nerf, the superstim needs a nerf, and their base health could also probably use a nerf
image
image
image
image

MatthewMarinets:
## Bulwark Company
Took a stab at a partial rebalance while I was here as it simplified the weapon somewhat.
* No longer benefits from Goliath Multi-Lock Weapons
* Level 1 perk changed. Used to be "anti-air attack now does full damage to all types", now gives blink
* Level 2 perk changed. Used to be anti-ground splash, now gives multi-lock weapons
* Reduced base health 250 -> 200
* Reduced health/level 70 -> 50
* Increased gas cost 325 -> 400
* Reduced AG damage 27 -> 20 (note: spartan is 24)
* Reduced AA damage 12+12a -> 11+11a (note: spartan is 11+11a)
image
image
image
image
image
image
image
image
image

Showing the multilock working (it took quite some fiddling, the 5 weapons visible in this picture were a debug measure that are not visible in the final commit)
image

MatthewMarinets:
## Sky Fury
* Fixed L3 shield effect
* Removing effect of Anti-Mechanical Munition
* Removing effect of Shredder Rounds
* Nerfed L3 shield effect life 400->300
* Nerfed L1 bonus damage to massive 50 -> 30

image
image
image
image
image
image
image
image

Fixing the L3 perk:
image
image

MatthewMarinets:
## Night Hawk
Didn't touch any balance numbers or upgrades. I think it would be cool to have them not benefit from advanced lasers and instead have level 3 let them shoot ground with their AA missiles (as the missiles are _way_ cooler than the laser), but will leave that for a future pass. I also just want to get this PR done.

### AA
image
image
image
image

### AG
image
image
image
image

## Anti-AG
image
image
image
image

MatthewMarinets:
## Liberator
* Mostly left infrastructure intact, as weapon swap was necessary for perks
* Eliminated unneeded behaviours for level 3, 4 AA and level 2, 3 AG and adjusted switches to only use the distinct weapons

### Balance
* changed level 3 perk from "can use AG attack vs air" to "can use AA attack while in AG mode" (which honestly is what I thought the perk was the first time I read it)
  * Note: I set the arc on the AA-while-in-AG-mode attack to 270, so they can't shoot behind themselves. Not setting it gave them a very narrow field of fire.
* Adjusted AG damage 125 -> 100
* Adjusted AA damage/missile 13 -> 10
* Adjusted Midnight Rider damage/missile 13 -> 12

## AA
image
image
image
image

## AG
image
image
image
image
image

PR #373, opened 2025-01-29
SC2: Fix Dark Whispers + Temple of Unification locations

Snarkie:
The locations to destroy the bases were able to trigger during the ending cinematics, even if you did not actually destroy the base.


MatthewMarinets:
Recommend reducing the trigger check period to 0.5s or 1s. Looks like 2s now, which is long enough for a player to start twiddling thumbs or starting to look for additional buildings.

Ziktofel:
The checking could be set to "Any unit dies" instead (if it isn't connected to some vanilla trigger)

Snarkie:
> The checking could be set to "Any unit dies" instead (if it isn't connected to some vanilla trigger)

I had some issue with the "unit dies" event in Dark Whispers. Periodic event works fine tho. Will change it to 0.5 seconds.
I use unit dies in Temple of Unification.

PR #376, opened 2025-02-06
sc2: Polish/balance adjustments (Trooper+Bile Launcher+Shield Battery)

MatthewMarinets:
2 changes based on feedback that's been pending a while:

## Dominion Trooper:
* -10 HP

This is to make them less of a "strictly better marine" at baseline (cheaper, faster). Applied this to all variants for consistency, as only the flamethrower should give bonus health.

## Bile Launcher
* -2 range
* No longer hits air
* -35 baseline damage (75 -> 40)
* Gets +30 bonus damage if built beside a spawning pool

There were comments that the bile launcher was somewhat OP on defense. There were some concerns about cost, DPS, and footprint. My hope with the spawning pool change is to leave small amounts of bile launchers viable / allow cheeky cheese bile launchers at forward positions, but nerf massed bile launchers by either taking a near 50% DPS cut or by packing them way less closely to make room for spawning pools (and also with a lower range, fewer can be placed to hit a choke). This also implicitly makes them more expensive mineral-wise.

## Shield Battery
Updated placement model to no longer have a weird 1-range CActorRange:
image

Old, pre-fix:
image

## Pictures
image
image
image
image

Note the two selected bile launchers don't get the buff, but the two unselected ones do. That is, the buff applies is a side length touches between the bile launcher/spawning pool footprints, but not diagonally or with a gap.
image

### DIFF
--- a/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
+++ b/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
@@ -2870,14 +2870,14 @@
         <BehaviorFlags index="Permanent" value="1"/>
         <InfoFlags index="Hidden" value="1"/>
         <BuffFlags index="DisableWhileUnderConstruction" value="1"/>
-        <Period value="0.25"/>
+        <Period value="2"/>
         <PeriodicEffect value="AP_BileLauncherBesideSpawningPoolAuraSearch"/>
     </CBehaviorBuff>
     <CBehaviorBuff id="AP_BileLauncherBesideSpawningPool">
         <InfoFlags index="Hidden" value="1"/>
         <EditorCategories value="Race:Zerg,AbilityorEffectType:Units"/>
         <Alignment value="Positive"/>
-        <Duration value="0.5"/>
+        <Duration value="2.5"/>
         <Modification>
             <DamageDealtScaled index="Splash" value="30"/>
         </Modification>
--- a/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
+++ b/Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
@@ -2865,6 +2865,23 @@
             <WeaponDisableArray value="AP_Suicide"/>
         </Modification>
     </CBehaviorBuff>
+    <CBehaviorBuff id="AP_SpawningPoolBileLauncherAura">
+        <EditorCategories value="AbilityorEffectType:Units,Race:Zerg"/>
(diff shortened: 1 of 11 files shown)


MindHawk:
Maybe there should be a visual indicator when a Bile Launcher is getting the spawning pool bonus

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
```diff
@@ -2865,6 +2865,23 @@
             <WeaponDisableArray value="AP_Suicide"/>
         </Modification>
     </CBehaviorBuff>
+    <CBehaviorBuff id="AP_SpawningPoolBileLauncherAura">
+        <EditorCategories value="AbilityorEffectType:Units,Race:Zerg"/>
+        <BehaviorFlags index="Permanent" value="1"/>
+        <InfoFlags index="Hidden" value="1"/>
+        <BuffFlags index="DisableWhileUnderConstruction" value="1"/>
+        <Period value="0.25"/>
```
MindHawk, line 2873: Could this period be adjusted up? A search per bile launcher is quite expensive and having it be a little unresponsive (like every 2s, have the buff last 3s) should not be too noticeable

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/EffectData.xml
```diff
@@ -13036,14 +13036,25 @@
...
     </CEffectEnumArea>
     <CEffectDamage id="AP_BileLauncherBombardmentDamage" parent="DU_WEAP">
         <EditorCategories value="Race:Zerg"/>
         <Kind value="Splash"/>
-        <Amount value="75"/>
+        <Amount value="40"/>
     </CEffectDamage>
+    <CEffectEnumArea id="AP_BileLauncherBesideSpawningPoolAuraSearch">
+        <EditorCategories value=""/>
+        <SearchFilters value="-;Self,Ally,Neutral,Enemy,Missile,Destructible,Dead,Hidden"/>
```
MindHawk, line 13049: Adding Structure to the required search filter would be slightly more performant iirc

MatthewMarinets: Done, added Structure and Biological tags.

Mods/ArchipelagoPlayer.SC2Mod/Base.SC2Data/GameData/BehaviorData.xml
MatthewMarinets: Upped the time to 2s period, 2.5 duration. Didn't add the switch as there's been talk of reworking how we do building unlocks for Yggdrasil/Warp Relocate reasons.

PR #391, opened 2025-02-22
SC2: Rework Hand of Darkness race swaps

Snarkie:

Added similar replacement units for the Zerg Brutalisk (as in big beefy units not obtainable otherwise):
 - Terran: ARES Warbot from Piercing the Shroud
 - Protoss: Stone Zealots from Smash & Grab

Rebalanced their stats to be closer to a Brutalisk, notably:

### Stone Zealot:
- 500hp -> 1000 hp + 500 shields. It was designed to be taken out by a handful of marines.
- buffed damage, removed detection (Brutalisk or Warbot don't get detection)
- Changed collision to Colossus, it felt weird to be clipped by Colossus but blocked by small units. 
- No change to cliff movement or vision, only collision

### Warbot:
- One rescued warbot gets all-purpose cannons, the other gets anti-armor missiles
- Adjusted hp and damage slightly, moved some of the armor bonus to baseline for anti-armor missiles
- Allowed anti-armor missiles to attack air, both Brutalisk and Stone Zealots have anti-air attacks
- Increased allied push priority, it was especially prone to getting blocked

### Changed Protoss starting units to be Tal'Darim themed:
- 2 Wrathwalkers, 1 Vanguard, 2 Havocs, 2 Slayers

Added some voicelines for the race swaps

## How was this tested?
Tested in local campaign, made sure the original race still works as well

## Graphical changes
Videos in discord channel


Ziktofel:
The Brutalisks are affected by AP W/A upgrades, these replacements should be too

Snarkie:
> The Brutalisks are affected by AP W/A upgrades, these replacements should be too

In the HotS mod, like the Brutalisk? Or in the map only? Are the units used in any other maps where this could have an impact?

Snarkie:
> The Brutalisks are affected by AP W/A upgrades, these replacements should be too

Fixed, decided on handling it in the map, just in case the units are used in other maps as well

Ziktofel:
Missing Quatro

Snarkie:
> Missing Quatro handling for Protoss

Fixed

PR #402, opened 2025-03-23
Raven Weapon Targeting Fix

EmptyJ:
Fixed Raven Weapon Targeting missiles, stasised and dead units.


Ziktofel:
What exactly is this supposed to fix? And why only targeting NCO (AP variant can exist outside of NCO)?

PR #406, opened 2025-03-31
Misc fixes

EmptyJ:
### SoA Guardian Shell
- Added Guardian Shell Behavior to non-protoss units
- Modified Requirement for Guardian Shell to trigger to also require a living AP_SOAAutonomousCaster to require enabled SOA Autocasting by proxy

### In The Enemy's Shadow
- Fix Pathable Terrain in North Secret Corridor

### For Aiur
- For Aiur Drones now auto-harvest resources.
- Enemy AI is now initialized.
- Enemy upgrade scaling is now functional as in Vanilla


EmptyJ:
I have verified there is no double-proc with zergling reconstitution and no proc when banelings/scourge attack something.

Ziktofel:
Check the SC2-next branch that has many of those fixes already applied

PR #409, opened 2025-04-04
Correct Two Overlayed Wireframes Displaying for Certain Zerg Strains 

EmptyJ:
Set strains that inherit wireframe from parents to instead override the inherited wireframe with the correct wireframe.


Ziktofel:
What's about the NCO region? Is that a vanilla bug? Can I see a screenshot?

EmptyJ:
I can upload a screenshot in about an hour, but it likely is in vanilla, but it doesn't really come up in vanilla because it is beyond where you need the blink suit and you don't have vision of the pathable terrain. You could likely come back from the gun blade room with the jumpsuit to reach it though.

EmptyJ:
image

Maps/ArchipelagoCampaign/NCO/ap_in_the_enemy_s_shadow.SC2Map/Objects
```diff
@@ -10492,6 +10492,7 @@
     <ObjectPoint Color="0,0,0,0" Id="1830833840" Name="No Fly Zone 054" PathingRadiusHard="2.5" PathingRadiusSoft="3" Position="105.42,141.4238,0" Scale="1,1,1" Type="NoFlyZone"/>
     <ObjectPoint Color="0,0,0,0" Id="1637971181" Name="No Fly Zone 034" PathingRadiusHard="2.5" PathingRadiusSoft="3" Position="49.741,174.5808,0" Scale="1,1,1" Type="NoFlyZone"/>
     <ObjectPoint Color="0,0,0,0" Id="1160309722" Name="S3 - Alarm Bot Patrol 1 - 1" Position="119,184.5,0" Scale="1,1,1" Type="Normal">
+    <ObjectPoint Id="2046855289" Position="97.051,224.5903,0" Scale="1,1,1" Type="BlockPathing" Name="Pathing Cliff Fill 057" Color="0,0,0,0"/>
         <Flag Index="ForcePlacement" Value="1"/>
     </ObjectPoint>
```
Ziktofel, line 10497: BTW check the XML syntax as you're creating an `ObjectPoint` within an another `ObjectPoint`

PR #420, opened 2025-04-14
The Great Train Robbery Raceswap Updates

SirChuckOfTheChuckles:
Generator PR:

Made the following changes:

### Zerg
Replaced Diamondbacks with Infested Diamonds. They deal slightly less DPS than regular diamondbacks, but they do have innate health regeneration and do gain increased movespeed on creep, so I considered it a fair trade. Adjusted some voicelines to account for the change in units.

### Protoss
Replaced Diamondbacks with Immortals. They are able to deal with the early trains easily; despite a slower movement speed they deal much more DPS to the trains and can take much more favorable trades with Hardened Shields. Adjusted some voicelines to account for the changes in units.

Also adjusted some in-game pings/icons/voicelines to account for the different races and their units.


Ziktofel:
Shouldn't be there Annihilators instead of Immortals for Protoss?

SirChuckOfTheChuckles:
There was some discussion in discord (link: https://discord.com/channels/731205301247803413/1154164338769268859/1358912909123653642) about whether or not it should be Immortals or Annihilators. While Annihilators are able to move while using Shadow Cannon, Immortals are able to easily have enough dps for the early trains. I also think they are thematic for the mission a bit more since they are good at taking split-up engagements with the attack waves/trains/kill teams, so I opted to include them instead of Annihilators.


Annihilators are also used in a bunch of raceswap missions already so I didn't feel a need to include them here for variety's sake.

PR #433, opened 2025-05-01
[SC2] Malignant Creep rework

Snarkie:
## What does this fix or add?

Moves the bonuses of Malignant Creep to a Behavior that is applied to all units which should be affected. This behavior affects the base regeneration and attackspeed of units instead of the regen/attack speed on creep, and it is disabled via a validator, while the unit is not on creep. 

This fixes several issues with units, most notably a permanent life drain bug. A previous, incomplete fix involving a marking behavior and a trigger is now obsolete. The trigger was changed to apply default behaviors instead, which will now apply the bonuses for Malignant Creep, should they be missing. The Malignant Creep upgrade no longer affects most of the units, it only enables the behavior, and upgrades the sight + cast range of Creep Tumors.

Additionally, Malignant Creep bonuses were added to the following units, that were not affected before:

* Infested Terrans of any variety, except the exploding Siege Tank ones (which were already affected)
  * this includes spawned variants by Infested Barracks, Infested Bunkers as well as Infestors
* Infested Buildings of any variety, except Infested Missile Turrets (which were already affected)
* Morphlings
* Infested SCV
* Burrowed Infested Banshees (flying ones are not affected, but burrowed ones do touch creep)

## Visual Changes

The buff behavior already existed ("AP_CreepFrenzy13Crawler") and was used by some units, I decided to re-use it instead of making a new behavior that does the exact same thing. However I did change the display name, and units currently affected by Malignant Creep bonuses now display a buff icon:
grafik
(The exact value of 0.8203 life per second has been rounded)

I decided to display the same tooltip for affected non-attacking units like most buildings as well. I prefer one consistent behavior, even if it displays an attack speed buff on buildings without a weapon.

**Any newly implemented unit,** which should be affected by Malignant Creep, **is required to add the "AP_CreepFrenzy13Crawler" behavior**. It is no longer required to add these units to an upgrade, or use a trigger upon acquiring these units.

## How was it tested?

Tested in local test maps, did extensive tests with single unit types to make sure all of the issues with current Malignant Creep are fixed. For specific cases covered, see discord channel.
Made a map with HotS campaign units to make sure story units like Stukov and Niadra are affected as well.


Ziktofel:
Check LotV mod (also used for Epilogue) as Stukov can be controllable there too (in Into the Void if take over AI allies is on)

Snarkie:
> Check LotV mod (also used for Epilogue) as Stukov can be controllable there too (in Into the Void if take over AI allies is on)

Fixed the LotV mod. Tested Into the Void Stukov and Essence of Eternity Zagara specifically.

Fixed certain campaign units getting double benefits. I missed the upgrades from the campaign dependencies.

PR #446, opened 2025-05-13
Yggdrasil Fixes/Functionality Updates

SirChuckOfTheChuckles:
The purpose of this PR is mainly to fix the bug with Yggdrasil where it is unable to properly place down buildings that the Player themselves has not unlocked yet. This was done through the following method:

**TECHSKIP buildings:** these are basically clones of the buildings, but just have an auto-morph ability similar to the mothership and spine/spore crawlers that morphs them to their respective building once construction is completely. This essentially allows the player to make these techskip buildings no matter what, even if the original building itself is not unlocked.

The yggdrasil then instead builds these techskip buildings, which then just becomes the normal buildings upon completion.

This was tested with each building that would require the techskip with Yggdrasil; 17, with the list below:

- Lair
- Hive
- Spawning Pool
- Roach Warren
- Baneling Nest
- Spine Crawler
- Spore Crawler
- Bile Launcher
- Evolution Chamber
- Ultralisk Cavern
- Infestation Pit
- Greater Spire
- Greater Hydralisk Den
- Hydralisk Den
- Spire
- Nydus Network
- Predator Nest

The original buildings were also checked with the actor changes, to make sure nothing went wrong in that department.

In the editor, the buildings were given a TECHSKIP - AP - prefix in order to distinguish them:

image

Also made the other changes to Yggdrasil:
Assimilate now has an 8-second cooldown, starting the moment the channel begins.
When assimilating a Hatchery, Lair, or Hive, Assimilate now destroys all nearby Larva.
Assimilate Functionality has been added for the Bile Launcher.


MatthewMarinets:
I don't think this is the best approach unless your primary goal is to avoid redoing unlock actions in the triggers (which I don't expect would take that long). This adds a whole bunch of dummy units, requires actor data changes, and adds morph abilities for what should really just be an alternate unlock condition.

The unlock condition can be implemented by either updating the requirements on the normal build abil to also require an upgrade (e.g. `AP_UnlockSpawningPool`), and unlocking the building in triggers by giving that upgrade; or by just updating all of the triggers to enable/disable build abils by ID rather than units.

PR #452, opened 2025-05-16
Building Trigger Rework

SirChuckOfTheChuckles:
This PR heavily revamps how buildings are unlocked/locked during a run by locking the ability command to make the building itself instead of the allowing/disallowing the actual building. The purpose of this is to fix the issues that are associated with Warp Relocate and Yggdrasil, where they are not able to build/relocate the building if the player has not unlocked it for building.

With this change, both Yggdrasil and Warp Relocate are able to properly move and build buildings the player has not even unlocked yet.

Some notes:
For most of the buildings, they can only be built by one ability.
The current exception to this rule is the static defensive structures that Dominion Troopers can build. This means that they need to be disabled/enabled for this unit as well, shown below:
image
image

Most of the trigger swaps were made in the clear{Race}Tech functions, unlocking units/staticdefense buildings, and then some certain upgrades that matter, such as:

The fusion core energy regen upgrade
Internalized Tech Module style-upgrades, which includes

- Medic
- Raven
- Cyclone
- Siege Tank
- Marauder
- Hercules
- Battlecruiser
- Wraith
- Banshee
- Goliath

Hive Mind Emulator/Psi Disruptor should work with all 3 variants
Tech Labs/Reactors should work properly for terran, and infested tech labs should work properly for zerg
Warp Harmonization should worker properly for Robotics Facilties/Stargates

Example of Before and After

Before:
image

After:
image


Ziktofel:
Mostly the cases are caused by improper use of `State="Restricted"` on some ability commands but this triggering should make stuff more robust to evil triggers. I'll remove it from Yggdrasil

Had to test a case if disabling/enabling one ability won't link somehow to the other one

PR #458, opened 2025-05-25
Merc Zerg Tooltips and Merc Roach Rename

Subsourian:
* Added flavor tooltips for the zerg mercs to match them with the terran mercs

* Renamed Thornshell to Caustic Horror as per discord discussion


Ziktofel:
I've updated the client

PR #465, opened 2025-05-29
SC2: Update Elder Probes

Snarkie:

After discussion on Discord, Elder Probes were considered fairly weak and not worth the gas investment. I wanted to buff them, without improving their impact on economy. This PR aims to improve their usability as a frontline probe:

- Elder Probes now provide a Power Field
  - this can be toggled on/off
- Elder Probes now automatically reconstruct on death at your Nexus
  - your Nexus get an autocast ability, that can reconstruct one Elder Probe every 20 seconds
  - this replaces the command card slot for the Elder Probe summon, as it is no longer needed after using it once
  - also uses the same hotkey
  - this ability gets 1 charge whenever an Elder Probe is destroyed
  - multiple Nexus have individual cooldowns, but share charges
  - you never get more than 5 Elder Probes at a time, but multiple Nexus allow you to reconstruct them faster
- Elder Probes get a slightly stronger weapon (5 > 7 damage) and are affected by Protoss Ground Weapon upgrades

----

- added custom icons for the abilities
- changed the Elder Probe handling trigger to use FindPlacement instead of spawning an additional Elder Probe (as that was messing with the reconstruction)

## How was this tested

Tested in local testmap. The client change only updates the description.

## Graphical changes
grafik
grafik grafik
grafik

grafik

[Client PR](https://github.com/Ziktofel/Archipelago/pull/474)


Ziktofel:
Inconsistency: Matrix overload doesn't work with Elder Probes (it does work in vanilla for Energizers and was also turned on for Warp Prisms in AP as the player doesn't have Warp Prisms in vanilla LotV)

Ziktofel:
BTW is there any reason for the player to turn off the power field?

Snarkie:
> BTW is there any reason for the player to turn off the power field?

Visual clutter, annoying to see moving power fields in your mineral line constantly

Snarkie:
> Inconsistency: Matrix Overload doesn't work with Elder Probes (it does work in vanilla for Energizers and was also turned on for Warp Prisms in AP as the player doesn't have Warp Prisms in vanilla LotV)

fixed

PR #478, opened 2025-06-04
SC2: Add Debug mode

Snarkie:

Adds a global Debug Mode trigger to the game
- If you use the command "-debug" on any mission, Debug Mode is enabled
- Debug Mode allows you to use additional debug commands

## Why is this needed?

When working on a mission, developers often implement commands to test specific features of the map. Many maps also come with already implemented debug commands, usually for skipping to a certain phase of the map.

Currently, each time a developer works on a map, they have to remember to deactivate/delete any debug commands they added/activated in the development process for the final version of the map. Accidentally leaving debug commands active in the map has caused issues on multiple occasions, especially since they are sometimes tied to button presses for comfortable access.

Making debug commands tied to a global, well-defined debug mode fixes this issue. Activating debug mode requires a deliberate action by the player, and it is clearly communicated they can now use special commands. This also allows developers to use their debug commands in the live version of the map without repercussions, which allows easier bug tracking and fixing.

## Details
grafik

Debug mode also allows for additional global commands. Currently implemented:
- "-control x": Takes control of player x (one-sided). Very helpful for debugging, especially for AI-related issues. Allows testing if the AI player is stopped by game mechanics, or if the actual AI behavior is at fault.

grafik

Example map: Enemy Intelligence
- "s2" or "s3": Pre-made debug command by Blizzard, skip to section 2 (no-build start) or section 3 (no-build half-way) of the map.

grafik
grafik

Debug Mode is currently NOT implemented as a YAML/client-side option. It is purely SC2-Data driven, enabled only with the "-debug" command. It can easily be added as a client option, but in my opinion, it is not necessary, as it is not intended to be used by the player.


Snarkie:
> I'd switch to `?Debug`

To be in line with the archipelago server messages? I don't like that at all. This is specifically not an AP server command, and ingame commands are usually done with -

PR #506, opened 2025-07-25
add general translations and map translations for german

Ekelalfred:
Added german translations back into maps and also the general data for all the buttons, tooltips, weapons etc
Most of the translations came from blizzards own translations and only the ones missing because of additions for AP were made by me in the style of blizzards translations


Ziktofel:
Also are all the current objectives there?

Ekelalfred:
Regarding the missing entries:
Not included entries will just show up as their keys. So the Skirmisher would show 'Unit/Name/AP_PhoenixTaldarim' whereever this string would normally be placed. And this is the main reason why I took the project up for me to translate all the keys missing in german. Because I worked with the general file for translation from the current master branch some keys are missing at the moment. I will correct this as fast as possible and also check if all the strings for the maps and objectives are still up to date or if I need to add some data over there too.
If this PR will be accepted and merged to the current dev-stage I would also offer to continue my work for this german translations as I can see that the development of this project is still ongoing and I want to support this process.

Ziktofel:
I plan to merge after all the current entries present in sc2-next will go in, so there won't be any degradation in UX. Also I plan to make a script that'd fill any missing string by its English variant

Ekelalfred:
I already did this for the big file in Player.SC2Mod and can push this if you want to merge this early. For all the maps I also did take a diff of what changed and need to be added

Ziktofel:
Also do the units play their (English) voices where their voices are backported or custom (Signifier, Medic, etc)?

Ekelalfred:
As far as I can tell the units all had voice lines at least in english. Sometimes it was a little mixed up but they were there.

Ziktofel:
@Ekelalfred Please check the report

Ekelalfred:
I saw your script output and will add these entries before the weekend.

Ziktofel:
The Infested Banshee's cloak tooltip got changed since you've updated it

PR #526, opened 2025-09-21
Bullfrog QoL changes

EnvyDragon:
- Improves jump validation, so situations where the unit would fail to move but the ability would go on cooldown should be less frequent
- Adds new ability: "All Aboard!" which orders nearby units to load into the Bullfrog, reducing APM tax


EnvyDragon:
-Fixed jump validator
-Fixed "All Aboard"
-Added a weak weapon: range 5, damage 8, period 1.8, slow projectile that can miss but has small AoE
-Raised cost to 125/75

